PR #5, opened 2023-10-04
Statechannels as an Escrow agent

NiloCK:
This PR demonstrates the general shape of using a state-channel network to manage escrow and payments for deals instead of a smart contract based system.

Proof of concept only - for educational purposes!


lukemarsden:
This is fascinating, I think @richardblythman might be interested in talking about this?

richardblythman:
Super interesting @NiloCK. Thanks for pointing me to this @lukemarsden.
Do we have a sense of the throughput and latency we could achieve with state channels vs L1?

PR #17, opened 2023-10-20
feat: template function subt

### LINKED ISSUE #14: Escaped json values are awkward
We json encode the string inputs to modules (i.e. the `-i` in `lilypad run cowsay:v0.0.1 -i Message=moo`) here: https://github.com/bacalhau-project/lilypad/blob/main/pkg/module/utils.go#L191-L199

Users are not trusted by nodes. Nodes should be able to trust module authors. If user input can make arbitrary changes to the job spec, this breaks the trust relationship.

Hence we want to avoid users doing nefarious things like including newlines and quotes in the untrusted user input.

This means that values in the json template, like https://github.com/bacalhau-project/lilypad-module-cowsay/blob/main/lilypad_module.json.tmpl#L16

like `{{.Message}}`, don't need to be escaped by the user with double quotes, but instead get automatically quoted for them. (Because the json serialization of a string has double quotes around it).

However, this makes it very awkward to substitute a value into an existing string, which many module authors want to do. In fact, we want to do it ourselves e.g. here: https://github.com/bacalhau-project/lilypad-module-sdxl/blob/main/lilypad_module.json.tmpl#L24

It would be much nicer to write
```
                    "PROMPT={{if .Prompt}}{{.Prompt}}{{else}}question mark floating in space{{end}}",
```
instead of 
```
                    {{if .PromptEnv}}{{.PromptEnv}}{{else}}"PROMPT=question mark floating in space"{{end}},
```
And then the user would be able to specify the prompt like `lilypad run sdxl:v0.9-lilypad1 -i Prompt="hoo haw"` instead of `lilypad run sdxl:v0.9-lilypad1 -i PromptEnv="PROMPT=hoo haw"` which is sad and annoying. But of course that would be insecure, because the user could write unquoted newlines and double quotes into PromptEnv and mess up the template.

To solve this, we need a way to securely substitute strings into other strings while still ensuring no JSON escape.

In `text/template`, which we use to parse the template you can define custom functions: https://pkg.go.dev/text/template#Template.Funcs

So,
- [ ] define a custom function which allows, given an untrusted user input string and, say, a printf style template string, for the module author to write:
```
                    {{subst "PROMPT=%s" .Prompt}},
```
Or something like that, where `subst` would:
* JSON decode `.Prompt` (we still want it encoded by default, for security reasons)
* printf it into the first given argument
* JSON encode the resulting string for security

Nasfame:
Resolves #14

### DIFF
--- a/pkg/module/utils.go
+++ b/pkg/module/utils.go
@@ -175,21 +175,21 @@ func PrepareModule(module data.ModuleConfig) (string, error) {
 
 func subst(format string, jsonEncodedInputs ...string) string {
 
-	jsonDecodedInputs := make([]string, len(jsonEncodedInputs))
+	jsonDecodedInputs := make([]any, 0, len(jsonEncodedInputs))
 
 	for _, input := range jsonEncodedInputs {
 		var s string
 
 		if err := json.Unmarshal([]byte(input), &s); err != nil {
-			log.Debug().AnErr("subst: json unmarshall", err).Msgf(input)
-			log.Printf("setting json encoded value for input:%s\n", input)
-			s = input
+			log.Debug().AnErr("subst: json unmarshall", err).Msgf("input:%s", input)
+			panic("subst: invalid input")
 		}
 
 		jsonDecodedInputs = append(jsonDecodedInputs, s)
 	}
+	log.Printf("jsonDecodedInputs:%v", jsonDecodedInputs)
 
-	return fmt.Sprintf(format, jsonDecodedInputs)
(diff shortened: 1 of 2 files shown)


lukemarsden:
Awesome! Thank you!!

lukemarsden:
Mind updating the readme as well please?

Nasfame:
@lukemarsden yes, ill do that. Currently working on a e2e test setup for the same

lukemarsden:
You are awesome! Thank you

PR #28, opened 2023-11-03
document cowsay on-chain; update HH .env with NETWORK=

hollygrimm:
Added scripts to `run-cowsay-onchain` and `runsdxl`

`hardhat/.env` now has a `NETWORK=` value. It defaults to `geth`


binocarlos (approved):
:+1:

PR #32, opened 2023-11-10
Update README.md to cover new multiarch builds

Zorlin:
Now that we have multiarch builds shipped, our instructions should reflect this

Also drops the Go compilation steps, although I'll re-add at least a reference to those in a new commit


DeveloperAlly (approved):
Could you also add these instructions in docs @Zorlin ?

Thankyou :)

PR #73, opened 2024-05-01
chore: update cowsay version smart contract call

AquiGorka:
These changes update the version in the docs and smart contract to trigger the `cowsay` module via a smart contract and as a freebie, remove some extra whitespace.


galaxyxtwo (approved):
Thank you @AquiGorka ! The update to the cowsay command should be sufficient to run with the new v3 commands

PR #85, opened 2024-05-14
README module edits, added lilysay

developersteve:


Zorlin (approved):

Zorlin (changes requested):
Please change this to use **sdxl-pipeline**, the old sdxl module is unsupported and broken

developersteve:
> Please change this to use **sdxl-pipeline**, the old sdxl module is unsupported and broken

Good catch, we probably need to look at depreciating/archiving and leaving notes on all the unsupported modules.

Zorlin:
> > Please change this to use **sdxl-pipeline**, the old sdxl module is unsupported and broken
> 
> Good catch, we probably need to look at depreciating/archiving and leaving notes on all the unsupported modules.

Deprecating (depreciating is an entirely different thing!), but yes! Definitely

Zorlin (approved):
LGTM

PR #102, opened 2024-05-20
refactor: update local dev stack and docs

bgins:

### Summary

This pull request updates the local dev stack and instructions with a few details uncovered while trying them out.

### Details (optional)

This PR implements the following changes:

- [x] Remove `lilypad` network
- [x] Add note to local development guide about default chain data path
- [x] Replace `yarn` with `npm` (our instructions ask the user to install node)


stack
```diff
@@ -5,6 +5,16 @@ IFS=$'\n\t'
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 OS_ARCH=$(uname -m | awk '{if ($0 ~ /arm64|aarch64/) print "arm64"; else if ($0 ~ /x86_64|amd64/) print "amd64"; else print "unsupported_arch"}')
 
+############################################################################
+# network
+############################################################################
+
+function network-create() {
```
AquiGorka: There is no need to specify the bridge, I think `bridge` is the default value, in any case `docker network create lilypad` should be good enough. Good catch!

bgins: It looks like we can remove the `lilypad` network entirely. Containers will use the default `bridge` network, which should work for our purposes unless we need an isolated network.

AquiGorka: Nice!
nit: are you familiar with [`git rebase interactive`](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)? I'd prefer this PR not have a commit that introduces a change that is later on removed on a subsequent commit, why go to that trouble if you can introduce the end changes in one commit without the back-and-forth.

bgins: Yep, makes sense! Dropped the unneeded commit. 👌

AquiGorka (approved):
🎉 🎉  thank you @bgins !

PR #111, opened 2024-05-23
change ExampleClient contract to use cowsay v0.0.4 (arm64 support)

richardbremner:
### Review Type Requested (choose one):
- [ x ] **Glance** - superficial check (from domain experts)

### Summary
change ExampleClient contract to use cowsay v0.0.4

### Task/Issue reference

Closes: [make cowsay available to all architectures](https://github.com/Lilypad-Tech/internal/issues/41)

### How to test this code? (optional)

When testing, running, need to make sure you remove the cloned, cached repository /tmp/lilypad/data/repos/lilypad-tech/lilypad-module-cowsay first. 

Also, need to make sure you have cleaned and run ./stack chain-boot, because the repo url (and version tag) is written into the smart contract.

PR #114, opened 2024-05-23
build: Add contracts compile and deploy job

### LINKED ISSUE #92: [core] Continuous deployment for smart contracts

bgins:

### Summary

This pull request adds a GH action job to compile and deploy smart contracts to devnet.

### Task/Issue reference

Closes: #92 

### Details (optional)

This pull request implements the following changes:

- [x] Add job to deploy contracts to devnet
- [x] Reduce the use of private keys when deploying contracts (only the admin key is needed)
- [x] Read RPC endpoint from environment

Note that this version of the job will deploy every time. We will want to check if re-deployment is necessary in a future PR.

We'll test this on `main` when we merge in #111, but a test run on this branch can be seen here: https://github.com/Lilypad-Tech/lilypad/actions/runs/9212958660/job/25345761697

### DIFF
--- a/.github/workflows/devnet_deploy.yml
+++ b/.github/workflows/devnet_deploy.yml
@@ -7,19 +7,25 @@ on:
 
 jobs:
   contracts:
-    runs-on: "ubuntu-latest"
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
           context: app
 
-      - name: Compile and deploy contracts
-        id: compile-and-deploy-contracts
+      - name: Install Doppler CLI
+        uses: dopplerhq/cli-action@v1
+
+      # TODO Avoid re-deploying contracts that are already on chain
+      - name: Deploy contracts
+        id: deploy-contracts
         env:
-          WEB3_RPC_URL: ${{ secrets.WEB3_RPC_URL }}
+          DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CONTRACTS_DEPLOY_DEVNET }}
         run: |
-          echo "# TODO"
+          cd hardhat
+          npm ci
+          doppler run -- npx hardhat deploy --network geth
 
   solver-build-deploy:
     runs-on: ubuntu-latest
--- a/hardhat/hardhat.config.ts
+++ b/hardhat/hardhat.config.ts
@@ -8,7 +8,7 @@ import * as dotenv from 'dotenv'
 
 import {
   ACCOUNT_ADDRESSES,
-  PRIVATE_KEYS,
+  getAccount,
 } from './utils/accounts'
 
 const ENV_FILE = process.env.DOTENV_CONFIG_PATH || '../.env'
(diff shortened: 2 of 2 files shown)


.github/workflows/devnet_deploy.yml
```diff
@@ -7,19 +7,25 @@ on:
...
+      - name: Deploy contracts
+        id: deploy-contracts
         env:
-          WEB3_RPC_URL: ${{ secrets.WEB3_RPC_URL }}
+          DOPPLER_TOKEN: ${{ secrets.DOPPLER_TOKEN_CONTRACTS_DEPLOY_DEVNET }}
         run: |
-          echo "# TODO"
+          cd hardhat
+          npm ci
+          doppler run -- npx hardhat deploy --network geth
```
bgins, line 28: Deploy implicitly compiles the contracts.

hardhat/hardhat.config.ts
```diff
@@ -27,16 +27,16 @@ const config: HardhatUserConfig = {
     geth: {
       url: 'http://localhost:8545',
       chainId: 1337,
-      accounts: PRIVATE_KEYS,
+      accounts: [getAccount('admin').privateKey]
```
AquiGorka, line 31: I have my suspicions that these may affect the development setup of the blockchain node.

Could you clear, run, fund and boot to verify this did not impact anything there?

hardhat/hardhat.config.ts
```diff
@@ -15,6 +15,7 @@ const ENV_FILE = process.env.DOTENV_CONFIG_PATH || '../.env'
 dotenv.config({ path: ENV_FILE })
 
 const NETWORK = process.env.NETWORK || "geth";
+const NETWORK_URL = process.env.WEB3_RPC_URL || 'http://localhost:8545';
```
AquiGorka: Can we give this a different name? From the way the env var `WEB3_RPC_URL` is used all around it points to the websocket endpoint (which for dev is `localhost:8546`), maybe `WEB3_HTTP_URL`?

bgins: Yeah good point. Updated it to `WEB3_HTTP_URL`.

hardhat/hardhat.config.ts
bgins: Just checked and run, fund, and boot for the chain all work with this change in local development.

The reason I think this is safe is the deploy scripts (https://github.com/Lilypad-Tech/lilypad/tree/main/hardhat/deploy) only deploy and excute commands from `admin`.

The public addresses for the other users appear to be configured here:

https://github.com/Lilypad-Tech/lilypad/blob/afbf4f8f9e45acca16615ffe730f19e2dfad2553/hardhat/hardhat.config.ts#L24

And they are referenced, but admin does all the work.

AquiGorka: Thank you for verifying.

AquiGorka (approved):
🎉 🎉  thank you @bgins !

PR #116, opened 2024-05-24
Get ENVS from Doppler for deployment

arsen3d:
Support to local_l2, arbitrum, polygon

**Logic** - thorough check (from everybody doing review)

### Summary
Easily deploy to any chain using doppler configuration.
Example:
./stack chain-deploy dev_personal

See dev_personal configuration under chain project in doppler

This enhancement should not effect any existing workflows.

### DIFF
--- a/hardhat/hardhat.config.ts
+++ b/hardhat/hardhat.config.ts
@@ -7,9 +7,25 @@ import 'hardhat-deploy'
 import * as dotenv from 'dotenv'
 
 import {
-  ACCOUNT_ADDRESSES,
-  PRIVATE_KEYS,
+  loadAddress,
+  loadPrivateKey,
+  getAllKeys,
+  setAccount,
+  getAllAddresses,
+  ACCOUNTS
 } from './utils/accounts'
+
+//override default publically known keys by doppler config
+for (const acc of ACCOUNTS) {
+  if (process.env[acc.name.toUpperCase()+'_PRIVATE_KEY'] && process.env[acc.name.toUpperCase()+'_ADDRESS']) {
+    acc.address = loadAddress(acc.name, process.env[acc.name+'_ADDRESS']?.toString() ?? '')
+    acc.privateKey = loadPrivateKey(acc.name, process.env[acc.name+'_PRIVATE_KEY']?.toString() ?? '')
+    setAccount(acc)
+  }
+}
+
+const PRIVATE_KEYS = getAllKeys()
+const ACCOUNT_ADDRESSES = getAllAddresses()
 
 const ENV_FILE = process.env.DOTENV_CONFIG_PATH || '../.env'
 dotenv.config({ path: ENV_FILE })
@@ -31,6 +47,36 @@ const config: HardhatUserConfig = {
     },
     sepolia: {
       url: `https://sepolia.infura.io/v3/${INFURA_KEY}`,
+      accounts: PRIVATE_KEYS,
+    },
+    local_l2: {
+      url: 'http://localhost:8547',
+      chainId: 412346,
+      accounts: PRIVATE_KEYS,
+    },
+    arbitrumSepolia: {
+      url: 'https://sepolia-rollup.arbitrum.io/rpc',
+      chainId: 421614,
+      accounts: PRIVATE_KEYS
+    },
+    arbitrumOne: {
+      chainId: 42161,
+      url: 'https://arb1.arbitrum.io/rpc',
+      accounts: PRIVATE_KEYS
+    },
+    arbitrumNova: {
+      chainId: 42170,
+      url: 'https://nova.arbitrum.io/rpc',
+      accounts: PRIVATE_KEYS
+    },
+    mumbai: {
+      chainId: 80001,
+      url: `https://polygon-testnet.public.blastapi.io`,
+      accounts: PRIVATE_KEYS,
+    },
+    polygon: {
+      chainId: 137,
+      url: `https://rpc.ankr.com/polygon`,
       accounts: PRIVATE_KEYS,
     },
   },
--- a/hardhat/utils/accounts.ts
+++ b/hardhat/utils/accounts.ts
@@ -84,3 +84,22 @@ export const getAccount = (name: string) => {
   }
   return account
 }
+export const getAllAddresses = () => {
+  return ACCOUNTS.reduce<Record<string, string>>((all, acc) => {
+    all[acc.name] = acc.address
+    return all
+  }, {})
+}
+export const getAllKeys = () => ACCOUNTS.map(acc => acc.privateKey)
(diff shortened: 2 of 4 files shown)


hardhat/hardhat.config.ts
```diff
@@ -34,6 +50,36 @@ const config: HardhatUserConfig = {
       url: `https://sepolia.infura.io/v3/${INFURA_KEY}`,
       accounts: [getAccount('admin').privateKey],
     },
+    local_l2: {
+      url: 'http://localhost:8547',
+      chainId: 412346,
+      accounts: PRIVATE_KEYS,
```
bgins, line 56: Is it necessary to pass all of the `PRIVATE_KEYS` in here? We found earlier in the deploy scripts that only `admin` deploys or executes initialization instructions. For example this works

https://github.com/Lilypad-Tech/lilypad/blob/14c3dc9432dbbeced4ce93aebd7ba1d244d4726c/hardhat/hardhat.config.ts#L47

It would be nice to reduce the number of private keys needed to run the deploy.

hardhat/hardhat.config.ts
```diff
@@ -7,10 +7,26 @@ import 'hardhat-deploy'
 import * as dotenv from 'dotenv'
 
 import {
-  ACCOUNT_ADDRESSES,
-  getAccount,
```
bgins, line 11: We are still using `getAccount`:

https://github.com/Lilypad-Tech/lilypad/blob/14c3dc9432dbbeced4ce93aebd7ba1d244d4726c/hardhat/hardhat.config.ts#L47
https://github.com/Lilypad-Tech/lilypad/blob/14c3dc9432dbbeced4ce93aebd7ba1d244d4726c/hardhat/hardhat.config.ts#L51

hardhat/hardhat.config.ts
```diff
@@ -7,10 +7,26 @@ import 'hardhat-deploy'
...
+  loadAddress,
+  loadPrivateKey,
+  getAllKeys,
+  setAccount,
+  getAllAddresses,
+  ACCOUNTS
 } from './utils/accounts'
 
+//override default publically known keys by doppler config
```
bgins, line 18:
```suggestion
//override default publicly known keys by doppler config
```

stack
```diff
@@ -343,5 +343,10 @@ function integration-tests() {
   cd test
   doppler run -p integration-tests -c dev -- go test -v -count 1 .
 }
+function chain-deploy() {
+  cd hardhat
+  echo $@
+  doppler run -p chain -c $@ -- yarn hardhat deploy
```
bgins, line 349:
```suggestion
  doppler run -p chain -c $@ -- npx hardhat deploy
```

Possible to switch to `npx`? Our `LOCAL_DEVELOPMENT.md` instructions ask users to install node, and it would be nice to avoid another tool.

hardhat/hardhat.config.ts
```diff
@@ -7,10 +7,26 @@ import 'hardhat-deploy'
...
+for (const acc of ACCOUNTS) {
+  if (process.env[acc.name.toUpperCase()+'_PRIVATE_KEY'] && process.env[acc.name.toUpperCase()+'_ADDRESS']) {
+    acc.address = loadAddress(acc.name, process.env[acc.name+'_ADDRESS']?.toString() ?? '')
+    acc.privateKey = loadPrivateKey(acc.name, process.env[acc.name+'_PRIVATE_KEY']?.toString() ?? '')
+    setAccount(acc)
+  }
+}
```
bgins, line 25: How does this improve the existing mechanism for loading from the environment? The `ACCOUNTS` constant calls `loadAddress` and `loadPrivateKey` when it is created, for example:

https://github.com/Lilypad-Tech/lilypad/blob/14c3dc9432dbbeced4ce93aebd7ba1d244d4726c/hardhat/utils/accounts.ts#L23-L28

hardhat/utils/accounts.ts
```diff
@@ -84,3 +84,22 @@ export const getAccount = (name: string) => {
...
+  if (foundAccount) {
+    console.log("account already exists, updating");
+    foundAccount.address = acc.address;
+    foundAccount.privateKey = acc.privateKey;
+    foundAccount.metadata = acc.metadata;
+    foundAccount.name = acc.name;
+  }
+  NAMED_ACCOUNTS[acc.name] = acc;
+  ACCOUNT_ADDRESSES[acc.name] = acc.address;
+};
```
bgins, line 105: nit: Could we use a single space between function definitions?

bgins (commented):
In general comment, it would be nice to only rely on the `admin` private key when deploying if we can. It reduces the number of keys we need to configure in Doppler and it helps us to reduce key exposure where we deploy.

arsen3d:
> In general comment, it would be nice to only rely on the `admin` private key when deploying if we can. It reduces the number of keys we need to configure in Doppler and it helps us to reduce key exposure where we deploy.

Yes. This is the case now. The other keys are hardcoded. They are known private keys. Those accounts will have the funds stolen if we don't generate new keys for them. We need to discuss how those keys are generated for production accounts, but this is independent of this pull request. I believe the deployment process funds other accounts from admin account.

bgins:
> I believe the deployment process funds other accounts from admin account.

Yes, that is my understanding as well.

> Yes. This is the case now. The other keys are hardcoded. They are known private keys. Those accounts will have the funds stolen if we don't generate new keys for them. We need to discuss how those keys are generated for production accounts, but this is independent of this pull request.

Agreed we need to consider a better mechanism for handling private keys in production. I think we can avoid being concerned with the non-admin private keys in this deploy script by using `getAccount('admin').privateKey` instead of `PRIVATE_KEYS` and associated mechanisms.

hardhat/utils/accounts.ts
```diff
@@ -84,3 +84,22 @@ export const getAccount = (name: string) => {
...
   return account
 }
+export const getAllAddresses = () => {
+  return ACCOUNTS.reduce<Record<string, string>>((all, acc) => {
+    all[acc.name] = acc.address
+    return all
+  }, {})
+}
+export const getAllKeys = () => ACCOUNTS.map(acc => acc.privateKey)
+export const setAccount = (acc: Account) => {
```
AquiGorka, line 94: Do we need to find the account? 

If you do:

```js
NAMED_ACCOUNTS[acc.name] = acc;
ACCOUNT_ADDRESSES[acc.name] = acc.address;
```

Then the mapped objects will hold the updated values

hardhat/hardhat.config.ts
AquiGorka: Oh wait, so `loadAddress` and `loadPrivateKey` already check for `env` vars? and the 2nd params is the default value? Then yeah, we don't need a new mechanism to inject `env` vars

Issue #79, opened 2024-05-07
[core] Check go installation of `lilypad` works


AquiGorka:
Reference: https://discord.com/channels/1212897693450641498/1237401839603814410/1237402224146124851

Zorlin:
```
wings:lilypad/ (main) $ GOPROXY=proxy.golang.org go list -m github.com/Lilypad-Tech/lilypad@latest                                                                                                                         [19:01:48]
github.com/Lilypad-Tech/lilypad v0.0.1-08ac6ae4e4e0776a0cab5dd04c4941a37cb00039
wings:lilypad/ (main) $ GOPROXY=proxy.golang.org go list -m github.com/Lilypad-Tech/lilypad/v2@latest                                                                                                                      [19:03:13]
go: module github.com/Lilypad-Tech/lilypad/v2: reading https://proxy.golang.org/github.com/%21lilypad-%21tech/lilypad/v2/@v/list: 404 Not Found
        server response: not found: github.com/Lilypad-Tech/lilypad/v2@v2.0.0-pre9: invalid version: go.mod has non-.../v2 module path "github.com/bacalhau-project/lilypad" (and .../v2/go.mod does not exist) at revision v2.0.0-pre9
```

We need to
* change the go.mod file to say github.com/Lilypad-Tech/lilypad/v2 instead of github.com/Lilypad-Tech/lilypad
* stop publishing tags as v2.0.0-gitshortshahash, and start publishing as v2.0.1, v2.1.0, etc, and follow Semantic Versioning correctly

richardbremner:
fixing up a whole bunch of "/v2" needing added through the codebase

bgins:
We updated the version to `v2`:

https://github.com/Lilypad-Tech/lilypad/blob/54076b20be701e0fa8a7ae1a474b73ec79dba9f0/go.mod#L1

but missed that the module path is case sensitive, so we'll also need to change `lilypad-tech` to `Lilypad-Tech`.

PR #132, opened 2024-05-31
change module version to v2

richardbremner:
- [ x ] **Glance** - superficial check (from domain experts)

### Summary
Upping the version to v2 and we'll use manual tagging to identify releases (for now)

### Task/Issue reference

[Closes: add_link_here](https://github.com/Lilypad-Tech/lilypad/issues/79)

### Details (optional)
We will leave the current circleci short sha tag versioning in place for now.

When we release we'll just create a tag with the next version number, ie.

git tag v2.0.x
git push origin v2.0.x

go install github.com/Lilypad-Tech/lilypad@latest will automatically pull down the highest semantic version


richardbremner:
fixing up a whole bunch of "/v2" needing added through the codebase

AquiGorka (approved):
@richardbremner approving to unblock, but I am not sure moving all the packages inside a `v2` namespace is the best thing to do, does it bring any benefit that we could put in place without such a big move?

AquiGorka:
@richardbremner is this PR still relevant?

richardbremner:
@AquiGorka i think it's still relevant, because it fixes the go install, which is otherwise still broken

AquiGorka:
If that's the case, then we should rebase, resolve the conflicts and merge.

AquiGorka:
@richardbremner let's figure out what we need to do here, the expectation is that after a fix is put in place we can do `go get github.com/Lilypad-Tech/lilypad/pkg/something/something` in a go codebase to install the required package (like, let's say, the [go bindings for the web3 contracts](https://github.com/Lilypad-Tech/lilypad/tree/main/pkg/web3/bindings) in the cron server, instead of copy pasting the abi and/or the binding file for pow)

richardbremner:
i think this is out of date and we can track those other things separately. at the time i think we still had `go install` in the docs and that wasn't working

PR #135, opened 2024-06-03
Add systemd watchdog support to service.

richardbremner:
- [ x ] **Logic** - thorough check (from everybody doing review)

### Summary
Uses the systemclt watchdog to automatically restart the Resource Provider service if it becomes unresponsive.

### Task/Issue reference

[Closes: add_link_here](https://github.com/Lilypad-Tech/infra/issues/2)

### Details (optional)
additional services can be included by specifying:
WatchdogSec=20s

in their .service file

systemctl daemon-reload
systemctl restart resource-provider

Monitor the notifications:
journalctl -u resource-provider -f

there should be a "Watchdog notification" message

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/cloudflare/circl v1.3.3 // indirect
 	github.com/consensys/bavard v0.1.13 // indirect
 	github.com/consensys/gnark-crypto v0.12.1 // indirect
+	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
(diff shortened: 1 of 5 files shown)


AquiGorka (approved):
Thank you @richardbremner 
Will this mechanism work with the AWS infra? One way to test for it is, how does this work when the resource provider is executed locally?

richardbremner:
This will work at the local machine level for each service.

At a level above that, we can creating cloudwatch rules that trigger based on the monitoring the metrics outputs (calls api endpoint) and takes certain actions based on the response.

AquiGorka:
Should we close this one? @richardbremner

PR #137, opened 2024-06-04
feat: Add network command line option and embedded configs

### LINKED ISSUE #131: [core] Update cli tool for incentivenet
- default points to "production" (which is going to be arbitrum testnet)
- identify the error when gcp is not around so that we can document it before people reach out to us
- add a `--devnet` flag that points to devnet (aws)
- maybe some way to auto upgrade the cli or maybe at least let people know they are not in the latest version (be mindful, not every single command, maybe once every N hours)

bgins:

### Summary

This pull request adds a global `--network` command line option that sets default services and web3 options for `testnet`, `devnet`, and `dev` networks. The defaults are pulled from configs embedded into the binary.

### Task/Issue reference

Closes: #131

### Details

The network defaults to `testnet` when the network flag is not used.

The embedded configs apply to `run`, `solver`,  `resource-provider`, `mediator`, and `jobcreator` commands. The following examples use the `run` command as an example.

#### Using Testnet defaults

```sh
go run . run cowsay:v0.0.1 -i Message=moo
```

This command uses the default values from `pkg/options/configs/testnet.toml`

#### Using Devnet defaults

```sh
go run . run cowsay:v0.0.1 -i Message=moo --network devnet
go run . run cowsay:v0.0.1 -i Message=moo -n devnet // shorthand version using -n
```

This command uses the default values from `pkg/options/configs/devnet.toml`

#### Override with an environment variable

```sh
WEB3_RPC_URL=http://example.com go run . run cowsay:v0.0.1 -i Message=moo
```

This command uses the default `testnet` values but overrides the RPC URL.

#### Override with a command line option

```sh
go run . run cowsay:v0.0.1 -i Message=moo --web3-chain-id 1337
```

This command uses the default `testnet` values but overrides the Chain ID.

It is also possible to override with environment variables and command line options at the same time.

### How to test this code?

The last commit on this PR is a temporary commit that prints the web3 environment for each of the commands. We will remove this commit after folks have had a chance to review.

Note that the chain IDs for `dev` and `devnet` are set to `421614` in preparation for our Arbitrum work, but we may need to override them to `1337` for the moment.

### DIFF
--- a/pkg/options/config.go
+++ b/pkg/options/config.go
@@ -1,0 +1,37 @@
+package options
+
+import (
+	"embed"
+	"errors"
+	"fmt"
+
+	"github.com/BurntSushi/toml"
+	"github.com/lilypad-tech/lilypad/pkg/data"
+	"github.com/lilypad-tech/lilypad/pkg/web3"
+)
+
+//go:embed configs
+var fs embed.FS
+
+type Config struct {
+	Web3          web3.Web3Options   `toml:"web3"`
+	ServiceConfig data.ServiceConfig `toml:"services"`
+}
+
+// TODO(bgins) Check for user-defined config files
+func getConfig(network string) (*Config, error) {
+	var config Config
+
+	path := fmt.Sprintf("configs/%s.toml", network)
+	config_toml, err := fs.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("network %s does not exist", network)
+	}
+
+	_, err = toml.Decode(string(config_toml), &config)
+	if err != nil {
+		return nil, errors.New("unable to parse config file")
+	}
+
+	return &config, nil
+}
--- a/pkg/options/configs/dev.toml
+++ b/pkg/options/configs/dev.toml
@@ -1,0 +1,14 @@
+[services]
+solver = "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC"
+mediator = ["0x90F79bf6EB2c4f870365E785982E1f101E93b906"]
+
+[web3]
+rpc_url = "ws://localhost:8546"
+chain_id = 421614
+controller_address = "0x4ed7c70F96B99c776995fB64377f0d4aB3B0e1C1"
+payments_address = "0x2279B7A0a67DB372996a5FaB50D91eAA73d2eBe6"
+storage_address = "0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e"
+users_address = "0x0DCd1Bf9A1b36cE34237eEaFef220932846BCD82"
+token_address = "0xa513E6E4b8f2a923D98304ec87F64353C4D5C853"
+mediation_address = "0x0B306BF915C4d645ff596e518fAf3F9669b97016"
+jobcreator_address = "0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE"
--- a/.github/workflows/devnet_deploy.yml
+++ b/.github/workflows/devnet_deploy.yml
@@ -61,6 +61,7 @@ jobs:
             --build-arg="expose_via=cloudflare" \
             --build-arg="cloudflare_token=${CLOUDFLARE_TOKEN_SOLVER}"\
             --build-arg="doppler_config=devnet" \
+            --build-arg="network=devnet" \
             .
           docker tag $ECR_REPOSITORY_SOLVER:latest $ECR_REGISTRY/$ECR_REPOSITORY_SOLVER:latest
           docker push $ECR_REGISTRY/$ECR_REPOSITORY_SOLVER:latest
@@ -122,6 +123,7 @@ jobs:
             -t $ECR_REPOSITORY_JOB_CREATOR \
             -f ./docker/job-creator/Dockerfile \
             --build-arg doppler_config=devnet \
+            --build-arg="network=devnet" \
             .
           docker tag $ECR_REPOSITORY_JOB_CREATOR:latest $ECR_REGISTRY/$ECR_REPOSITORY_JOB_CREATOR:latest
           docker push $ECR_REGISTRY/$ECR_REPOSITORY_JOB_CREATOR:latest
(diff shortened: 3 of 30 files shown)


pkg/options/config.go
```diff
@@ -0,0 +1,36 @@
...
+type Config struct {
+	Web3          web3.Web3Options   `toml:"web3"`
+	ServiceConfig data.ServiceConfig `toml:"services"`
+}
+
+// TODO(bgins) Check for user-defined config files
+func getConfig(network string) (*Config, error) {
+	var config Config
+
+	path := fmt.Sprintf("configs/%s.toml", network)
```
AquiGorka, line 25: Curious as to why you went with `fmt.Sprintf` to concatenate, it seems to be the slowest method ([ref](https://dev.to/schadokar/the-fastest-way-to-concatenate-strings-in-golang-5mf))

AquiGorka:
We should modify the `stack` commands to run the services to include the `network` param set to `dev` (unless I tested this wrong but it seems the env var for `NETWORK` did not override the selected network)

AquiGorka:
These changes make it so that we don't need doppler to run the local dev network right? That's a great thing for us to be able to receive open source contributions.

We should update the local development document.

AquiGorka:
nit (at the time of this PR) to test locally the `chainId` in the configs for `dev` and `devnet` has to be set to `1337`

pkg/options/configs/dev.toml
```diff
@@ -0,0 +1,14 @@
+[services]
+solver = "0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC"
+mediator = ["0x90F79bf6EB2c4f870365E785982E1f101E93b906"]
```
AquiGorka, line 3: Because I don't really know how to *use* the mediator, I can't really validate this it the right address, but we can come back to it later on when we have more context on the validator.

pkg/options/config.go
```diff
@@ -0,0 +1,36 @@
...
+	config_toml, err := fs.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("Network %s does not exist", network)
+	}
+
+	_, err = toml.Decode(string(config_toml), &config)
+	if err != nil {
+		return nil, fmt.Errorf("Unable to parse config file")
```
AquiGorka: nit: is this idiomatic? Returning a `fmt.Errorf`? Maybe we could do `fmt.Error` or `Error`? No real action here, just want to land on a convention

AquiGorka (approved):
🎉 🎉 great work @bgins !
I tested for `dev` and `devnet`, but the real wild ride (testing on `testnet`) I leave to you.

pkg/options/configs/dev.toml
bgins: Sounds good. FWIW, these were the values we had in Doppler in the `dev` environment.

pkg/options/config.go
bgins: Did a quick bit of reading, and I have a proposal for a convention:

- `errors.New` when the error is only a string and does not need formatting
- `fmt.Errorf` when we want formatting
- Custom errors when we want to add extra data fields for context (i.e. not just a string)

Surprisingly, `fmt.Error` does not exist! Perhaps an indication that we are encouraged to use `errors.New`.

Following that convention, I would update these errors to use `errors.New`.

How does that sound? Some references that brought me to this proposal:

- https://www.digitalocean.com/community/tutorials/handling-errors-in-go
- https://www.reddit.com/r/golang/comments/6ffrie/fmterrorf_or_errorsnew/

AquiGorka: Thank you for the follow up and research, let's follow the conventions you outlined.

AquiGorka: Quick follow up: it seems errors in go as a convention are lowercased (most prob so that devs can do `fmt.Println("Something happened:", err)`)

pkg/options/config.go
bgins: In a word, naivety. 😁 

But reading up on it a bit, I think it's an acceptable choice for this use case:

- We aren't in a tight loop or situation where we are constructing strings that often. We'll do this once per command run.
- It's more readable that using the `+` operator. We know immediately what we are doing, without considering operands to `+`.
- `Strings.builder` would take multiple lines of code and would be less readable

Open to the other options, but for code readability this seems like the best option. Thoughts?

bgins:
> We should modify the `stack` commands to run the services to include the `network` param set to `dev` (unless I tested this wrong but it seems the env var for `NETWORK` did not override the selected network)

Yep, will do! Do we want a `NETWORK` environment variable? Could add that, but would have to rename these:

https://github.com/Lilypad-Tech/lilypad/blob/d63d8b5a668f80087ed28040947920041d3fdf70/hardhat/hardhat.config.ts#L17-L18

> These changes make it so that we don't need doppler to run the local dev network right? That's a great thing for us to be able to receive open source contributions.
> 
> We should update the local development document.

We still need the private keys that are stored in Doppler, but perhaps we update the local development docs with instructions to override the Doppler environment? Would need to research and experiment a bit, so maybe not on this PR.

> nit (at the time of this PR) to test locally the `chainId` in the configs for `dev` and `devnet` has to be set to `1337`

Good catch! Will update that. 💯

> 🎉 🎉 great work @bgins ! I tested for `dev` and `devnet`, but the real wild ride (testing on `testnet`) I leave to you.

pkg/options/config.go
bgins: Ah yeah! Will update that 👌

pkg/options/config.go
AquiGorka: Your reasoning works for me

AquiGorka:
>  Do we want a NETWORK environment variable?

No, no, I was just thinking out loud. An update to the stack commands is good enough.

AquiGorka:
> We still need the private keys that are stored in Doppler, but perhaps we update the local development docs with instructions to override the Doppler environment? Would need to research and experiment a bit, so maybe not on this PR.

True. We'll punt on this, but this PR open that door is a good step in that direction

PR #138, opened 2024-06-04
feat: Add GCP deprecation notice

bgins:

### Summary

Adds a check to see if the user is attempting to use the GCP Testnet.

### Task/Issue reference

Implements #131 

### Details

The check is triggered when the GCP RPC endpoint is used. It fatally stops the program and instructs the user how to migrate to the new Testnet.

For example, when the chain ID and GCP RPC endpoint are set:

```
This testnet has been deprecated. Please remove all environment variables for RPC URL, chain ID, and contract addresses.

Your environment currently contains:

(web3.Web3Options) {
 RpcURL: (string) (len=30) "ws://testnet.lilypad.tech:8546",
 PrivateKey: (string) (len=5) "*****",
 ChainID: (int) 1337,
 ControllerAddress: (string) "",
 PaymentsAddress: (string) "",
 StorageAddress: (string) "",
 UsersAddress: (string) "",
 TokenAddress: (string) "",
 MediationAddress: (string) "",
 JobCreatorAddress: (string) "",
 Service: (system.Service) (len=11) "job-creator"
}

(data.ServiceConfig) {
 Solver: (string) "",
 Mediator: ([]string) <nil>
}

The new testnet environment will be set by default, and you will only need to set your private key.
```

### How to test this code?

Run the `run` command in any environment that uses the old endpoint. This can be forced with a command line option:

```sh
go run . run cowsay:v0.0.1 -i Message=moo --web3-rpc-url ws://testnet.lilypad.tech:8546
```

### DIFF
--- a/cmd/lilypad/solver.go
+++ b/cmd/lilypad/solver.go
@@ -3,6 +3,7 @@ package lilypad
 import (
 	"fmt"
 
+	"github.com/lilypad-tech/lilypad/pkg/data"
 	optionsfactory "github.com/lilypad-tech/lilypad/pkg/options"
 	"github.com/lilypad-tech/lilypad/pkg/solver"
 	memorystore "github.com/lilypad-tech/lilypad/pkg/solver/store/memory"
@@ -20,6 +21,8 @@ func newSolverCmd() *cobra.Command {
 		Long:    "Start the lilypad solver service.",
 		Example: "",
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			optionsfactory.CheckDeprecation(data.ServiceConfig{}, options.Web3)
+
 			network, _ := cmd.Flags().GetString("network")
 			options, err := optionsfactory.ProcessSolverOptions(options, network)
 
--- a/cmd/lilypad/jobcreator.go
+++ b/cmd/lilypad/jobcreator.go
@@ -19,6 +19,8 @@ func newJobCreatorCmd() *cobra.Command {
 		Long:    "Start the lilypad job creator service.",
 		Example: "",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			optionsfactory.CheckDeprecation(options.Offer.Services, options.Web3)
+
 			network, _ := cmd.Flags().GetString("network")
 			options, err := optionsfactory.ProcessOnChainJobCreatorOptions(options, args, network)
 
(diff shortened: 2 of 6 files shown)


cmd/lilypad/solver.go
```diff
@@ -20,6 +21,8 @@ func newSolverCmd() *cobra.Command {
 		Long:    "Start the lilypad solver service.",
 		Example: "",
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			optionsfactory.CheckDeprecation(data.ServiceConfig{}, options.Web3)
```
bgins, line 24: The solver does not have a service config.

AquiGorka: It doesn't? Then we should remove the `SERVICE_MEDIATORS` key/value pair from the `solver`'s environments in doppler.

AquiGorka (approved):
🍬 (sweet)
Before you merge can you clear a confusion I have: why do you some times check `options.Offer.Services` and one other time `options.Services`? I couldn't figure this one out

cmd/lilypad/solver.go
bgins: Yep agreed. To back this up, the solver options are here:

https://github.com/Lilypad-Tech/lilypad/blob/d5f16cadbed8f9b8223192357f9ccf9eb068d522/pkg/options/solver.go#L9-L16

The commands that configure services call `GetDefaultServicesOptions`:

https://github.com/Lilypad-Tech/lilypad/blob/d5f16cadbed8f9b8223192357f9ccf9eb068d522/pkg/options/services.go#L10-L15

bgins:
> 🍬 (sweet) Before you merge can you clear a confusion I have: why do you some times check `options.Offer.Services` and one other time `options.Services`? I couldn't figure this one out

The placement of `Services` is different depending on the command that is run. For example, in the `mediator` it is at the top level:

https://github.com/Lilypad-Tech/lilypad/blob/d5f16cadbed8f9b8223192357f9ccf9eb068d522/pkg/options/mediator.go#L11-L19

The `job-creator` has it nested one level deep. At the top level we have `Offer`:

https://github.com/Lilypad-Tech/lilypad/blob/d5f16cadbed8f9b8223192357f9ccf9eb068d522/pkg/options/job-creator.go#L12-L20

And `Offer` has `Services`:

https://github.com/Lilypad-Tech/lilypad/blob/d5f16cadbed8f9b8223192357f9ccf9eb068d522/pkg/options/job-creator.go#L28-L38

PR #139, opened 2024-06-05
install cron and schedule calling the pow process of lillypad

richardbremner:
- [ x ] **Logic** - thorough check (from everybody doing review)

### Summary
changes the docker contain to install cron, and create a schedule for initiating the proof-of-work

### Task/Issue reference

https://github.com/Lilypad-Tech/internal/issues/72

### Details (optional)
Not ready to be merged in yet because the "lilypad pow" command doesn't exist yet

### DIFF
--- a/docker/resource-provider/Dockerfile
+++ b/docker/resource-provider/Dockerfile
@@ -12,16 +12,38 @@ RUN go install
 
 RUN (curl -Ls --tlsv1.2 --proto "=https" --retry 3 https://cli.doppler.com/install.sh || wget -t 3 -qO- https://cli.doppler.com/install.sh) | sh
 
+# Install cron, used for PoW scheduler
+RUN apt-get update && apt-get install -y cron
+
+# TODO: introduce some form of randomisation here
+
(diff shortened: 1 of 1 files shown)


AquiGorka (commented):
i love the elegant solution proposed here, but the `Dockerfile` for resource providers has not been finalized and maybe users may not go with the docker way of running this, they may execute the `resource-provider` and `bacalhau` services directly.

If we integrate the cron execution directly into the `resource-provider` code, the solution would work for either way of setting up a node.

Side note: It was my understanding that we had agreed on an execution that was synced for all nodes (i'm not talking about the one we trigger) but something that goes more in the lines of "all nodes should execute pow at 12, then 12:30, then 13, etc" - to make it so that every node has the same window of time to submit their results and with your code here, the cron will execute every 30 minutes after the service starts, which may not make all nodes to execute in sync.

AquiGorka:
@richardbremner should we close this one?

richardbremner:
cancelled, no longer doing this

PR #145, opened 2024-06-06
feat: add tracking to resource provider

### LINKED ISSUE #134: [core] Figure out on the Resource Provider code where we can get data about the hardware it provides and track data
We want to trigger calls to the api `/metrics-dashboard/nodes` endpoint (websocket) with accompanying data about the nodes, ideally:

- Node identifier (maybe a mix of gpu specs + public key + location + platform (os and architecture))
- User identifier (public key, not used as primary because more than one RP node can be using the same private key)
- Hardware specs (if possible, and this is something we need to figure out how to get, maybe bacalhau has some cmd that returns some info)
- Location (if possible, we can start with geolocation by ip (yes, I know it is unreliable but could be a good first approach), cloudflare may offer some [geolocation service](https://github.com/Lilypad-Tech/internal/issues/48))
- These could come with the connect event for a node, we also want to track disconnect events, so we need the node code to spin up a websockets server back to us so that we can track connect/disconnect events.
- Jobs received
- Jobs state (the job itself success/fail)
- Job time to return (either success or fail)

Alongside this task, we need to add a web sockets server endpoint to the api where nodes will connect and send the data. Then insert this data into a new table: `nodes` (id, node id, user, specs, location, connected_timestamp, disconnected_timestamp) <- we will have more than one record per node so that we can add up all the times the node was "online".
During the active connection, the node can send the jobs updates via the websocket or call an endpoint (whichever is simplest).

AquiGorka:

### Summary
These changes:
- Add a mechanism to trigger events when the websocket connection is in place and when it disconnects
- Add a payload to the websocket connection to identify the type of service connecting and the public address from it
- Use the mechanism from above to update the number of resource providers connected to the solver
- Use the mechanism from above to track connect/disconnect events
- Track node info when a resource provider connects to the solver

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/lilypad/issues/134

Follow the instructions here to test: https://github.com/Lilypad-Tech/api/pull/40

### DIFF
--- a/pkg/http/types.go
+++ b/pkg/http/types.go
@@ -7,6 +7,8 @@ type ServerOptions struct {
 }
 
 type ClientOptions struct {
-	URL        string
-	PrivateKey string
+	URL           string
+	PrivateKey    string
+	PublicAddress string
+	Type          string
 }
--- a/pkg/http/websocket_server.go
+++ b/pkg/http/websocket_server.go
@@ -21,12 +21,19 @@ type ConnectionWrapper struct {
 	mu   sync.Mutex
 }
 
+type WSConnectionParams struct {
+	ID   string
+	Type string
+}
+
 // StartWebSocketServer starts a WebSocket server
 func StartWebSocketServer(
 	r *mux.Router,
 	path string,
 	messageChan chan []byte,
 	ctx context.Context,
+	connectCB func(params WSConnectionParams),
+	disconnectCB func(params WSConnectionParams),
 ) {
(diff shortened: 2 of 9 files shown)


bgins (approved):
Looks good! One detail, perhaps the PR title should use `feat`? It's not a user-facing feature, but it does add functionality.

logan-wrld (commented):
awesome to see how the /controller(s) are used here

PR #150, opened 2024-06-10
limit the containers that can be executed

### LINKED ISSUE #63: [core] Limit containers that can be executed
Every official node should follow the allowlist defined in https://github.com/orgs/Lilypad-Tech/projects/4/views/2?pane=issue&itemId=60261819

logan-wrld:

### Summary
Limit the containers that can be executed on the jobcreator. Updates consist of a module allowlist fetch and allowlist check before deal agreement. 

### Task/Issue reference
Closes: #63

### Details 

Implemented into the jobcreator/controller.go due to controller managing job offers and doing the interaction with the solver on deal agreements. This includes the allowlist fetch using a simple http fetch to the lilypad-tech Github as well as the allowlist check function before a deal is agreed on. 

I am still having issues with my lilypad on-chain jobs in my local dev environment- please test as well!

### Anything else? (optional)
I did a alias to the http package as to not step on the "github.com/lilypad-tech/lilypad/pkg/http" package


pkg/jobcreator/controller.go
```diff
@@ -190,6 +220,56 @@ func (controller *JobCreatorController) Start(ctx context.Context, cm *system.Cl
 
 	return errorChan
 }
+func (controller *JobCreatorController) UpdateModuleAllowlist() error {
+	// URL of the GitHub raw content of the allowlist
+	allowlistUrl := "https://raw.githubusercontent.com/lilypad-tech/module-allowlist/master/allowlist.txt"
+
+	// Create a new HTTP request with a context
+	req, err := httpAlias.NewRequest("GET", allowlistUrl, nil)
```
bgins: We might want to add a new helper function to the `"github.com/lilypad-tech/lilypad/pkg/http"` package to handle this case and others that need a timeout. We could create a `GetRequestWithTimeout` that takes the necessary request parameters and a timeout argument.

The idea is to keep all of our HTTP communication going through one package. And it's pretty likely we'll need something like this again in the future!

pkg/jobcreator/controller.go
```diff
@@ -190,6 +220,56 @@ func (controller *JobCreatorController) Start(ctx context.Context, cm *system.Cl
...
+	// Update the internal cache of the allowlist
+	controller.moduleAllowlist = strings.Split(strings.TrimSpace(string(body)), "\n")
+	return nil
+}
+
+func (controller *JobCreatorController) allowlistApproved() error {
+	controller.web3Events.Storage.SubscribeDealStateChange(func(ev storage.StorageDealStateChange) {
+		deal, err := controller.solverClient.GetDeal(ev.DealId)
+		if err != nil {
+			controller.log.Error("module allolist error", err)
```
bgins:
```suggestion
			controller.log.Error("module allowlist error", err)
```

pkg/jobcreator/controller.go
```diff
@@ -190,6 +220,56 @@ func (controller *JobCreatorController) Start(ctx context.Context, cm *system.Cl
 
 	return errorChan
 }
+func (controller *JobCreatorController) UpdateModuleAllowlist() error {
+	// URL of the GitHub raw content of the allowlist
+	allowlistUrl := "https://raw.githubusercontent.com/lilypad-tech/module-allowlist/master/allowlist.txt"
```
bgins: Could we make this a package level const? If we won't change it at runtime, we could have it as a read-only value.

Also, is this URL correct? Currently reports a 404.

pkg/jobcreator/controller.go
```diff
@@ -145,8 +149,29 @@ func (controller *JobCreatorController) subscribeToWeb3() error {
 }
 
 func (controller *JobCreatorController) Start(ctx context.Context, cm *system.CleanupManager) chan error {
+	// Initial fetch of the module allowlist
+	err := controller.UpdateModuleAllowlist()
+	if err != nil {
+		// handle error
```
bgins: How do we plan to handle this error?

logan-wrld: I fixed this by passing the error to the controller with contoller.log.Error... Does this work? Especially since this error is mainly for testing, passing it to web3 metrics is overkill?

pkg/jobcreator/controller.go
logan-wrld: Makes way more sense than a hardcoded string! The allowlistURL is now sourced from a package-level constant instead of being hard-coded within the function. The module-allowlist is now configured on Github so you should not be seeing a 404 error any longer.

pkg/jobcreator/controller.go
logan-wrld: I added a GetRequestWithTimeout helper function within the pkg/http/utils.go - HTTP comms are going through one package!

bgins: Awesome! Thanks ❤️

pkg/jobcreator/controller.go
bgins: Yeah, agreed passing it to metrics probably isn't needed. We'd probably get a bunch of reports whenever someone typos cowsay. 😆

Let's chat about the `controller.log.Error` approach. It's probably right, but would love to chat about it because I am still getting familiar with the code base.

pkg/jobcreator/controller.go
```diff
@@ -228,6 +290,7 @@ func (controller *JobCreatorController) solve() error {
  *
 */
 
+// list the deals we have been assigned to that we have not yet posted and agree tx to the contract for
```
bgins, line 301: We can remove this one. It's a duplicate of the line below.

logan-wrld:
Closed in favor or issue #166

PR #165, opened 2024-06-17
fix: devnet config

AquiGorka:

### Summary
Apparently the `Devnet` blockchain node does not generate the same addresses as the local node.

- Run bacalhau `./stack bacalhau-node`
- Connect an RP `doppler run --preserve-env -p resource-provider -c devnet -- go run . resource-provider "$@" --network devnet`
- Run a job `doppler run --preserve-env -p run -c devnet -- go run . run cowsay:v0.0.4 -i Message="Devnet" --network devnet`
- Validate the job was succesful and the local RP shows output for the job


bgins (approved):
Nice catch, thanks for updating these. 🙏

PR #166, opened 2024-06-17
feat: module allowlist

### LINKED ISSUE #63: [core] Limit containers that can be executed

logan-wrld:

### Summary
Limit the containers that can be executed on the jobcreator. Updates consist of a module allowlist fetch and allowlist check before deal agreement. 

### Task/Issue reference
Closes: #63

### Details 

 **Location of Module Allowlist**: https://github.com/Lilypad-Tech/module-allowlist/blob/main/allowlist.txt

Implemented into the jobcreator/controller.go due to controller managing job offers and doing the interaction with the solver on deal agreements. This includes the allowlist fetch using a simple http fetch to the lilypad-tech Github as well as the allowlist check function before a deal is agreed on. 

I am still having issues with my lilypad on-chain jobs in my local dev environment- please test as well!

### Anything else? (optional)
I did a alias to the http package as to not step on the "github.com/lilypad-tech/lilypad/pkg/http" package


richardbremner:
So we are checking this https://raw.githubusercontent.com/lilypad-tech/module-allowlist/main/allowlist.txt list every 5 seconds? 

Maybe I misunderstood, but that seems odd to me, I assume there's something I don't understand?

logan-wrld:
@richardbremner The code initializes a one-time fetch of the module allowlist when the controller starts and then sets up a periodic update via a time.Ticker that triggers once every hour

pkg/http/utils.go
```diff
@@ -397,3 +398,31 @@ func newRetryClient() *retryablehttp.Client {
...
+func GetRequestWithTimeout(url string, timeout time.Duration) ([]byte, error) {
+	// A HTTP client with the specified timeout
+	client := &http.Client{
+		Timeout: timeout,
+	}
+
+	// Create the HTTP request
+	req, err := http.NewRequest("GET", url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating module-allowlist request failed: %s", err)
```
AquiGorka: I think this error is specific to the allowlist feature but the intention is for this client to be generic no? My suggestion here just do `creating request failed` and the caller will prepend with the appropriate `return fmt.Errorf("failed to fetch module allowlist: %s", err)`

pkg/http/utils.go
```diff
@@ -397,3 +398,31 @@ func newRetryClient() *retryablehttp.Client {
...
+	}
+
+	// Perform the HTTP request
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("executing module-allowlist request failed: %s", err)
```
AquiGorka: Same comment as above, let's do `return nil, fmt.Errorf("executing request failed: %s", err)`

pkg/http/utils.go
```diff
@@ -397,3 +398,31 @@ func newRetryClient() *retryablehttp.Client {
...
+	}
+	defer resp.Body.Close()
+
+	// Read the response body
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading module-allowlist response body failed: %s", err)
```
AquiGorka: One more, let's change this to `return nil, fmt.Errorf("reading response body failed: %s", err)`

pkg/jobcreator/controller.go
```diff
@@ -25,13 +26,18 @@ type JobCreatorController struct {
...
+	moduleAllowlist       []string
 }
 
-// the background "even if we have not heard of an event" loop
-// i.e. things will not wait 10 seconds - the control loop
-// reacts to events in the system - this 10 second background
-// loop is just for in case we miss any events
-const CONTROL_LOOP_INTERVAL = 10 * time.Second
+const (
+	allowlistURL = "https://raw.githubusercontent.com/lilypad-tech/module-allowlist/main/allowlist.txt"
```
AquiGorka: [nit] Why is this one not in all caps? I may be off too, do let me know if there's some convention here.

pkg/jobcreator/controller.go
```diff
@@ -287,6 +349,7 @@ func (controller *JobCreatorController) agreeToDeals() error {
 // list the deals that have results posted but we have not yet checked
 // we do this synchronously to prevent us racing with large result sets
 // also we are the client so have a lower chance of there being a chunky backlog
+
```
AquiGorka, line 364: Any reason for this extra line here?

pkg/http/utils.go
logan-wrld: Thanks- Totally makes sense!

pkg/jobcreator/controller.go
logan-wrld: I updated `allowlistURL` to `ALLOWLIST_URL` to match the control loop const. I agree- better to match the convention and be neat!

pkg/jobcreator/controller.go
logan-wrld: sorry about that!

10d9e (approved):
LGTM

LOCAL_DEVELOPMENT.md
```diff
@@ -79,3 +79,5 @@ There are two commands that can be used to run existing tests: `./stack unit-tes
 ## Notes on tooling
 
 Things should work right out-of-the-box, no extra configuration should be needed as Doppler provides the environment variables that are required with the current setup.
+
+ß
```
AquiGorka, line 83: This seems off, let's remove this.

pkg/data/types.go
```diff
@@ -140,7 +145,8 @@ type ResourceOffer struct {
 	Spec MachineSpec `json:"spec"`
 	// the module ID's that this resource provider can run
 	// an empty list means ALL modules
-	Modules []string `json:"modules"`
+	Modules []EnabledModules `json:"modules"`
```
AquiGorka, line 148: Can you rename this to `EnabledModules`? `Modules` feels too generic and the intention may get lost in other places where this is used.

pkg/jobcreator/controller.go
```diff
@@ -141,6 +148,7 @@ func (controller *JobCreatorController) subscribeToWeb3() error {
 	controller.web3Events.Storage.SubscribeDealStateChange(func(ev storage.StorageDealStateChange) {
 		deal, err := controller.solverClient.GetDeal(ev.DealId)
 		if err != nil {
+			err = fmt.Errorf("module allowlist error: %s", err)
```
AquiGorka, line 151: This looks off, if that's the error then it should bubble up from where it happened no? The way this code reads it says any error during the `GetDeal` process is a module allowlist err

pkg/resourceprovider/resourceprovider.go
```diff
@@ -60,11 +60,18 @@ type ResourceProviderPowOptions struct {
...
 type ResourceProviderOptions struct {
-	Bacalhau bacalhau.BacalhauExecutorOptions
-	Offers   ResourceProviderOfferOptions
-	Web3     web3.Web3Options
-	Pow      ResourceProviderPowOptions
+	Bacalhau  bacalhau.BacalhauExecutorOptions
+	Offers    ResourceProviderOfferOptions
+	Web3      web3.Web3Options
+	Pow       ResourceProviderPowOptions
+	Allowlist ResourceProviderAllowlistOptions
```
AquiGorka, line 74: Same here, let's deprecate `allowList` and bring forth `enabledModules`

pkg/solver/matcher.go
```diff
@@ -18,58 +19,311 @@ func (a ListOfResourceOffers) Less(i, j int) bool {
...
+		var allowlist []AllowlistItem
+		if err := json.Unmarshal(bytes, &allowlist); err != nil {
+			return nil, err
+		}
+	
+		return allowlist, nil
+	}
+
+	// strip out https- keep the org and module in place
+	// https://lilypad.tech/cowsay -> lilypad.tech/cowsay
```
AquiGorka, line 49: It is more like `https://github.com/Lilypad-Tech/lilypad-module-cowsay` -> `lilypad-tech/lilypad-module-cowsay` (take notice how they are all lower case)

pkg/solver/matcher.go
```diff
@@ -18,58 +19,311 @@ func (a ListOfResourceOffers) Less(i, j int) bool {
...
+	// Assume module ID has the shape of the org/repo(eg lilypad-tech/cowsay) if string
+	func moduleMatch(moduleID, allowlistModuleID string) bool {
+		return moduleID == allowlistModuleID
+	}
+	
+	func versionMatch(version, allowlistVersion string) bool {
```
AquiGorka, line 55: an we get unit tests for different pattern matching? Here's two references that we can rely on:
- https://docs.npmjs.com/cli/v6/using-npm/semver
- https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies

pkg/solver/matcher.go
```diff
@@ -18,58 +19,311 @@ func (a ListOfResourceOffers) Less(i, j int) bool {
...
+
+/*// if the job offer is zero then it will match any resource offer
 func doOffersMatch(
 	resourceOffer data.ResourceOffer,
 	jobOffer data.JobOffer,
-) bool {
-	if resourceOffer.Spec.CPU < jobOffer.Spec.CPU {
+) bool */
+func doOffersMatch(resourceOffer data.ResourceOffer, jobOffer data.JobOffer, allowlist []AllowlistItem, allowlistEnabled bool) bool {
+	if !allowlistEnabled {
```
AquiGorka, line 73: Does this mean: "if the allow list feature is not enabled" then all offers match?

AquiGorka:
Hey @noryev it looks like these changes still include comments on TODOs, can you double check you've pushed all the changes and let us know?

PR #172, opened 2024-06-19
fix: don't log metrics calls when API_HOST is empty

### LINKED ISSUE #171: [core] If the API_HOST env var is not set, do not trigger calls in metrics dashboard
In development we may not always have the `api` server running and we get an error in stdout (`Post "http://localhost:8002/metrics-dashboard/deals": dial tcp [::1]:8002: connect: connection refused`) from the calls to track events, let's disable triggering those calls if the host is not set ([reference](https://github.com/Lilypad-Tech/lilypad/blob/main/pkg/metricsDashboard/logger.go#L19))

walkah:

### Summary

This PR adds a guard in `trackEvent` to skip the http call if the `API_HOST` environment variable isn't set. 

This also includes a little bit of re-factoring: made `trackEvent` private (all calls go through `TrackEventName` ... e.g. `TrackJobOfferUpdate`), and centralized the url string inside trackEvent. Let me know if that part is a step too far! (if there were plans, e.g. to re-use `TrackEvent` in other places).

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/lilypad/issues/171

- Make sure API_HOST is unset (doppler dev seems to set it to http://localhost:8002) - e.g. `export API_HOST=""`
- Do a stack run, e.g. `./stack run cowsay:v0.0.4 -i Message="yay`

You should no longer see "Post "http://localhost:8002/metrics-dashboard/deals": dial tcp [::1]:8002: connect: connection refused"

### DIFF
--- a/pkg/metricsDashboard/logger.go
+++ b/pkg/metricsDashboard/logger.go
@@ -16,9 +16,14 @@ const nodeInfoEndpoint = "nodes"
 const nodeConnectionEndpoint = "uptimes"
 const dealsEndpoint = "deals"
 
-var host = os.Getenv("API_HOST") + "metrics-dashboard/"
+func trackEvent(path string, json string) {
+	var host = os.Getenv("API_HOST")
+	if host == "" {
+		return
+	}
 
-func TrackEvent(url string, json string) {
+	var url = host + "metrics-dashboard/" + path
+
 	data := []byte(json)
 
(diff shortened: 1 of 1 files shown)


AquiGorka (approved):
🎉 🎉 thank you @walkah !
and good catch on making `trackEvent` private!

PR #155, opened 2024-06-11
build:one dockerfile for resource provider and bacalhau

taoshengshi:
### Review Type Requested (choose one):
- [x] **Glance** - superficial check (from domain experts)
- [ ] **Logic** - thorough check (from everybody doing review)

### Summary
The is the implementation of https://github.com/Lilypad-Tech/lilypad/issues/93 to give an option to end users and local dev, to pack these resource provider and bacalhau services together.

The background of idea is to explore one approach to making Doppler optional in the Dockerfile so that we can set env vars at the run command: `docker run -e xx=yy` and for doppler env file: `docker run -d --env-file <(doppler secrets download --no-file --format docker) <image>`

### Task/Issue reference

Closes: #93 

### Details (optional)
#93 

### How to test this code? (optional)
1.  Build image:
`docker build -t <image name> .`

2. Run image:
- 2.1 docker run -e WEB3_PRIVATE_KEY= \<yy\> -it \<image\>
- 2.2 docker run -d --env-file \<(doppler secrets download --no-file --format docker) \<image\>

3. A `WEB3_PRIVATE_KEY` environment variable must be provided at run time
- 3.1 Get the `WEB3_PRIVATE_KEY` from the secrets provided by Doppler
```
docker run -d \
--env-file <(doppler secrets download --no-file --format docker) \
-v local/path:/tmp/lilypad/data \
<image name>
```
- 3.2 Provide `WEB3_PRIVATE_KEY` directly
```
docker run -d \
-e WEB3_PRIVATE_KEY=<Your Private Key> \
-v local/path:/tmp/lilypad/data \
<image name>
```
We set initial values for these environment variables:
```
ENV LOG_TYPE=json
ENV LOG_LEVEL=debug
ENV HOME=/app/lilypad
ENV BACALHAU_SERVE_IPFS_PATH=/data/ipfs
ENV OFFER_GPU=1
```

but you can reset them at run time if necessary.
```
docker run -d \
-e WEB3_PRIVATE_KEY=<Your Private Key> \
-e OFFER_GPU=2 \
-v local/path:/tmp/lilypad/data \
<image name>
```

### How to test the 'docker -v' to mount the volume of the host system with the container ?

1. Specify the data in the current folder as the data volume mount path
```
docker run -d \
--env-file <(doppler secrets download --no-file --format docker) \
-v $(pwd)/data:/tmp/lilypad/data \
demo
```
2. Successfully mounted
```
docker volume ls
DRIVER    VOLUME NAME
local     data
[...]

PR #175, opened 2024-06-20
fix: check combined outputs for potential errors when getting bacalhau job id

AquiGorka:

### Summary
These changes check the combined output from `bacalhau create --id-only --wait path/to/job/spec` for errors because I noticed that sometimes when the `id` was returned, the rest of the output also contained an error and the code was not checking for it, the rest of the process would execute but there would be no results in the output folder.

I don't think I have a straightforward way to reproduce the error, I noticed the situation when testing this [PR](https://github.com/Lilypad-Tech/lilypad/pull/155), I added a print statement to the `runOutput` var and I noticed the full output also included info to the lines of `could not execute job, not enough nodes` (I'm writing this from memory so if you are testing this bear in mind it may be a bit different) and the whole cli run was executed but the output folder did not include any files.
With these changes the process did return an error in that situation and the RP console showed the error had happened.

### DIFF
--- a/pkg/executor/bacalhau/bacalhau.go
+++ b/pkg/executor/bacalhau/bacalhau.go
@@ -128,9 +128,16 @@ func (executor *BacalhauExecutor) getJobID(
 	)
 	runCmd.Env = executor.bacalhauEnv
 
-	runOutput, err := runCmd.CombinedOutput()
+	runOutputRaw, err := runCmd.CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("error running command %s -> %s, %s", deal.ID, err.Error(), runOutput)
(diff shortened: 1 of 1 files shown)


walkah (approved):
🚢 Looks good to me! If it's tricky to reproduce, perhaps worth a unit test to verify? (obviously that won't catch if bacalhau changes the output formatting...)

AquiGorka:
> If it's tricky to reproduce, perhaps worth a unit test to verify?

That's a great suggestion, tracking [here](https://github.com/Lilypad-Tech/lilypad/issues/176) because I need to focus on another task but I do want to get my hands dirty with unit tests - we have been relying on existing tests and have not written news tests ourselves.

>  obviously that won't catch if bacalhau changes the output formatting

We are pinning `bacalhau@1.3.2` for now because their next release will bring breaking changes.

PR #192, opened 2024-06-26
fix: show bacalhau id error

hunjixin:

show error in "bacalhau id" command.

### Summary

### Task/Issue reference

Re: https://github.com/Lilypad-Tech/hni/issues/7

### Details (optional)

### DIFF
--- a/pkg/executor/bacalhau/bacalhau.go
+++ b/pkg/executor/bacalhau/bacalhau.go
@@ -51,11 +51,17 @@ func (executor *BacalhauExecutor) Id() (string, error) {
 
 	runOutputRaw, err := nodeIdCmd.CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("error calling get id results %s", err.Error())
+		return "", fmt.Errorf("error calling get id results %s, %s", err.Error(), runOutputRaw)
 	}
 
-	splitOutputs := strings.Split(strings.Trim(string(runOutputRaw), " \t\n"), "\n")
+	splitOutputs := strings.Split(string(runOutputRaw), "\n")
 	runOutput := splitOutputs[len(splitOutputs)-1]
+	outputError := strings.Join(strings.Fields(strings.Join(splitOutputs[:len(splitOutputs)-1], " ")), " ")
+	if outputError != "" {
+		// TODO: we need to figure out if errors here are fatal or not
+		log.Warn().Msgf("error calling bacalhau id: %s", outputError)
+	}
+
 	var idResult struct {
 		ID       string
 		ClientID string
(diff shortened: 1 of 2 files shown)


AquiGorka (commented):
@hunjixin thank you for the changes, can you provide a video recording of how this changes affect the POW process?

hunjixin:
bacalhau id have error
image

and when runing rp, lilypad not print pow worker message
image

this pr return this error, and tell user you got sometheing error
image

@AquiGorka

hunjixin:
image

add a new commit to resolve this problem,  cmd output some odd error log. just parse the last line  @AquiGorka

pkg/executor/bacalhau/bacalhau.go
```diff
@@ -49,18 +49,20 @@ func (executor *BacalhauExecutor) Id() (string, error) {
...
 	nodeIdCmd.Env = executor.bacalhauEnv
 
-	output, err := nodeIdCmd.CombinedOutput()
+	runOutputRaw, err := nodeIdCmd.CombinedOutput()
 	if err != nil {
 		return "", fmt.Errorf("error calling get id results %s", err.Error())
 	}
 
+	splitOutputs := strings.Split(strings.Trim(string(runOutputRaw), " \t\n"), "\n")
+	runOutput := splitOutputs[len(splitOutputs)-1]
```
AquiGorka, line 58:
```go
errOutput = splitOutputs[1:]
if errOutput != nil {
  return "", fmt.Errorf("error unmarshalling job JSON %s %s", err.Error(), errOutput)
}
```

hunjixin: but in correct situation,  [1:] may panic
image

hunjixin: image
add like this?, but i got a panic in this change. i dont know how the error log look like

AquiGorka: If you are getting errors, it means there is something wrong, you also need to trim, I added a similar change recently, here: https://github.com/Lilypad-Tech/lilypad/pull/175/files
Let's make sure we also capture errors if there are errors

walkah: @AquiGorka I've updated to add logging for error output.
@hunjixin Looks like we want `splitOutputs[:len(splitOutputs)-1]` (i.e. everything but the last line).

walkah:
In discussing with @taoshengshi - if we can get this merged, he can do testing in the morning (china time) with RPs to confirm this unblocks the issue.

AquiGorka (approved):

pkg/executor/bacalhau/bacalhau.go
hunjixin: this change seem to make things not work.the output in my comment 
1 line: error log
1 line: id result
1 line: empty line

this is why add a trim to make sure the empty line is removed

split by \n,  4 lines, with  2 empty line in last

PR #194, opened 2024-06-26
fix: tracking job offer update events

AquiGorka:

### Summary
These changes make it so that tracking the job offer update is not a subscription to job offer updates but a method call right before the subscriptions are executed.

### Details (optional)
Why?
I wish I knew why the previous way was not working, locally it worked, so the assumption was that it would also work in the deployed infra. 
I have a hypothesis the previous subscription was part of the client code and because the client *should* not have access to the `API_HOST` env var the requests should not have been going out, but as it happens, the doppler settings for `run` did include the `API_HOST` - although, after removing it, locally the events would be still tracked 🤷 .

- Setup a local network
- Spin up the api server
- Trigger both a cli jobn and an onchain job, then run the integration tests
- With each execution verify the `jobs` table in the db has the corresponding job entry (in the api repo:`./stack psql`, then `SELECT * FROM jobs;`)

### DIFF
--- a/pkg/metricsDashboard/logger.go
+++ b/pkg/metricsDashboard/logger.go
@@ -16,8 +16,9 @@ const nodeInfoEndpoint = "nodes"
 const nodeConnectionEndpoint = "uptimes"
 const dealsEndpoint = "deals"
 
+var host = os.Getenv("API_HOST")
+
 func trackEvent(path string, json string) {
-	var host = os.Getenv("API_HOST")
 	if host == "" {
 		return
 	}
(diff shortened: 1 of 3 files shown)


pkg/metricsDashboard/logger.go
```diff
@@ -16,8 +16,9 @@ const nodeInfoEndpoint = "nodes"
 const nodeConnectionEndpoint = "uptimes"
 const dealsEndpoint = "deals"
 
+var host = os.Getenv("API_HOST")
```
AquiGorka, line 19: @walkah I moved this out of the function call, I don't think we need to call it every time the function executes right?

walkah: You're absolutely right! Good catch 👍

walkah (approved):
🚀 lgtm!

PR #211, opened 2024-06-28
Arsen/build deploy pow

arsen3d:
- [x ] **Glance** - superficial check (from domain experts)

### Summary
build deploy pow


.github/workflows/devnet_deploy_contracts.yml
```diff
@@ -0,0 +1,26 @@
+name: Deploy Testnet contracts
```
AquiGorka, line 1: Do we really need this? Why can't we rely on the resetting the node every time? The intention for devnet is to test out infrastructure and we should be able to provision it fast and simple.

AquiGorka:
Closing this PR as we've already merged in the work for contract deployment state and deploying POW as a proxy.

PR #212, opened 2024-06-28
fix: pow proxy deployment

arsen3d:
- [x ] **Logic** - thorough check (from everybody doing review)

### Summary
Deployment of  POW Proxy Contract


richardbremner (approved):
lgtm

.github/workflows/testnet_deploy_contracts.yml
```diff
@@ -21,4 +21,6 @@ jobs:
         run: |
           cd hardhat
           npm ci
-          doppler run -- npx hardhat deploy --network arbitrumSepolia
+          doppler run -- npx hardhat deploy --network testnet
+          git commit -m "Contract Migration Updated"
```
AquiGorka, line 25: @arsen3d  @richardbremner is this the correct way to go here? If I recall correctly we've had to setup tokens to be able to do this in other workflows. Would this trigger deploys? Is that what we want? Can we get a clear description of what these changes are doing? I don't think:

```md
Summary
Deployment of POW Proxy Contract
```

Really expresses the full impact these changes have.

AquiGorka: cc @bgins would this get us in problems? Some sort of deploy loop?

bgins: Discussed this briefly with @AquiGorka. We think this could be redesigned a bit.

We think this workflow would be used in this context:

- A contract change is made on a branch and a PR opened
- The `testnet_deploy_contracts.yml` is run manually on the branch (select the branch from the dropdown)
- On success, the new contract addresses are added to the embedded configs and committed to the branch
- On failure, fix the issue and run again

This keeps changes neatly on a single PR and keeps the deployment, migrations, and embedded configs in sync.

We should verify this checks out the branch where `workflow_dispatch` is run:

https://github.com/Lilypad-Tech/lilypad/blob/bf1e8e4a46455d46d221f66db55e0bafe176fe47/.github/workflows/testnet_deploy_contracts.yml#L9-L10

Most likely it does but good to check.

We should make sure this workflow cannot run on `main` because is protected, and we can't commit directly to it. We could do this with an `if` conditional (https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution) on the entire `contracts` job.

We should `git add` the files before this `git commit`:

https://github.com/Lilypad-Tech/lilypad/blob/bf1e8e4a46455d46d221f66db55e0bafe176fe47/.github/workflows/testnet_deploy_contracts.yml#L25

Also, let's update the commit message to use a conventional commit. We can use `chore: Contract Migration Updated`. Maybe `chore` isn't totally ideal, but this will be on a branch with a `feat` or `fix` so it seems alright to hardcode this one to chore.

Lastly, for error handling we might separate the git add and commit into a separate step. That way if the deploy steps fails, we won't continue on and commit anything (this might be safe as is, but a separate step would make it more explicit)

AquiGorka: Potentially, if it can be done, it would be a source of great confidence on the upgrade if we could do a dry run with devnet.

Issue #195, opened 2024-06-26
Update Bacalhau to use RPC interface (instead of shell commands)


walkah:
Bacalhau > v1.4 deprecated the commands that we use to run jobs:  https://docs.bacalhau.org/references/cli-reference/command-migration

Also, using the CLI to run jobs requires bacalhau installed locally (which means bundling bacalhau inside our RP docker container).

hunjixin:
bacalhau has a lib to do this. things became more stable with this api.

https://github.com/bacalhau-project/bacalhau/tree/main/pkg/publicapi/client

bgins:
We should consider removing the Bacalhau executor integration out of `﻿pkg/data` into `pkg/executor/bacalhau` alongside the executor. At the time of writing, we also use Bacalhau for it's embedded IPFS, so some parts of the code may be more appropriate in `directory` or `storage` package.

PR #214, opened 2024-06-28
fix: pow issue

hunjixin:

1. revert bacalhau id 
2. remove parse event
### Summary

### Task/Issue reference

Closes: add_link_here

### Details (optional)

### DIFF
--- a/pkg/executor/bacalhau/bacalhau.go
+++ b/pkg/executor/bacalhau/bacalhau.go
@@ -54,13 +54,8 @@ func (executor *BacalhauExecutor) Id() (string, error) {
 		return "", fmt.Errorf("error calling get id results %s, %s", err.Error(), runOutputRaw)
 	}
 
-	splitOutputs := strings.Split(string(runOutputRaw), "\n")
+	splitOutputs := strings.Split(strings.Trim(string(runOutputRaw), " \t\n"), "\n")
 	runOutput := splitOutputs[len(splitOutputs)-1]
-	outputError := strings.Join(strings.Fields(strings.Join(splitOutputs[:len(splitOutputs)-1], " ")), " ")
-	if outputError != "" {
-		// TODO: we need to figure out if errors here are fatal or not
-		log.Warn().Msgf("error calling bacalhau id: %s", outputError)
-	}
 
 	var idResult struct {
 		ID       string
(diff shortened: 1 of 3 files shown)


pkg/executor/bacalhau/bacalhau.go
```diff
@@ -54,13 +54,8 @@ func (executor *BacalhauExecutor) Id() (string, error) {
 		return "", fmt.Errorf("error calling get id results %s, %s", err.Error(), runOutputRaw)
 	}
 
-	splitOutputs := strings.Split(string(runOutputRaw), "\n")
+	splitOutputs := strings.Split(strings.Trim(string(runOutputRaw), " \t\n"), "\n")
 	runOutput := splitOutputs[len(splitOutputs)-1]
-	outputError := strings.Join(strings.Fields(strings.Join(splitOutputs[:len(splitOutputs)-1], " ")), " ")
```
AquiGorka, line 59: Why are you removing this output? We cannot just silently disregard a potential error message in the combined output

hunjixin: the id command ouput
udp buf size error log
result json.


this code not work.
maybe we can check by line number,if lines more than one. give a warn log. do you think it okay?

AquiGorka: How can we fix the udp buf size error?

hunjixin: udp buf eeror occur in bacalau id command. i dont know how to fix.
maybe change some system config can fix that.

AquiGorka: Alright, capture the id, disregard the rest.

taoshengshi: for the udp buffer error, this is from bacalhau document:

> While executing this command, you may encounter warnings regarding receive and send buffer sizes: failed to sufficiently increase receive buffer size. These warnings can arise due to limitations in the UDP buffer used by Bacalhau to process tasks. Additional information can be found in https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes.

I tested it and it works!

richardbremner: I tested this too and it seems to work.

I think it's ok to discard the error output here because Bacalhau seems to be including irrelevant error output in the id command output. That's weird behavior.

richardbremner (approved):
lgtm

pkg/executor/bacalhau/bacalhau.go
hunjixin: i think we should change to use rpc instead of subprocess command.

https://github.com/Lilypad-Tech/lilypad/issues/195

PR #215, opened 2024-06-29
fix: json unmarshalling error bacalhau id

likhita-8091:
json content is empty, so error


richardbremner:
thanks for the contribution @CodingJzy , but there's another fix for this in a different PR https://github.com/Lilypad-Tech/lilypad/pull/214 that we're going to merge.

fwiw, I think there's a few issue with this PR:
1. bacalhau seems to sometimes include irrelevant error output with the `id` command before the json object, and this code didn't seem to work with that
2. the error output is not a fatal condition, I think we can safely ignore it

PR #217, opened 2024-07-01
chore: exit early if the current version is not available

richardbremner:
- [ x ] **Glance** - superficial check (from domain experts)

### Summary
I just added a check that the current version is available, because otherwise the script could do weird things

### DIFF
--- a/ops/lilypad-updater-gpu.sh
+++ b/ops/lilypad-updater-gpu.sh
@@ -13,12 +13,17 @@ LATEST_URL=$(curl -s https://api.github.com/repos/lilypad-tech/lilypad/releases/
 
 # Extract the latest version (including short SHA) from the URL
 LATEST_VERSION=$(echo $LATEST_URL | sed -n 's#.*/download/\([^/]*\)/.*#\1#p')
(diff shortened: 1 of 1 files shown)


ops/lilypad-updater.sh
```diff
@@ -13,12 +13,17 @@ LATEST_URL=$(curl -s https://api.github.com/repos/lilypad-tech/lilypad/releases/
...
 
 # Get the current version
 CURRENT_VERSION=$(/usr/local/bin/lilypad version | grep "Lilypad:" | awk '{print $2}')
-
 echo "Current version: $CURRENT_VERSION"
-echo "Latest version: $LATEST_VERSION"
+
+# Check if CURRENT_VERSION has a value
+if [ -z "$CURRENT_VERSION" ]; then
+    echo "Error: Unable to determin CURRENT_VERSION."
```
bgins:
```suggestion
    echo "Error: Unable to determine CURRENT_VERSION."
```

bgins (approved):
Looks good, left a comment with a typo fix.

Issue #89, opened 2024-05-16
[core] Add integration tests to CI for lilypad


AquiGorka:
If anything at least a single test that can run a whole job to prevent regressions from being released.

AquiGorka:
We could also add a couple of processes to the cron server (when it is implemented) that trigger jobs every N and alert us if the job was not successful.

bgins:
Tasks:

- [ ] Add GH action to run unit and integration tests on pull request to `main`
- [ ] Make sure integration tests use embedded configs

AquiGorka:
I was reading up on [kind](https://kind.sigs.k8s.io/) and it really surprised me how easily it could be used to setup a cluster for testing purposes - this is their claim too:

> [kind](https://sigs.k8s.io/kind) is a tool for running local Kubernetes clusters using Docker container “nodes”.
kind was primarily designed for testing Kubernetes itself, but may be used for local development or CI.

Also [Minikube](https://minikube.sigs.k8s.io/docs/tutorials/continuous_integration/) looks like could be another good choice

richardbremner:
@AquiGorka you also might want to have a look at docker swarms. https://docs.docker.com/engine/swarm/

AquiGorka:
> Use Swarm mode if you intend to use Swarm as a production runtime environment.

Not sure if that would be what we want for a CI step

walkah:
I've used `minikube` in the past, but looking into `kind` and doing some experiments yesterday - I think it's the way to go (and seems to have the community momentum as well). 

Related, I'm starting to put together some helm charts to spin up various configurations: https://github.com/Lilypad-Tech/helm-charts

bgins:
Scoped this issue to focus only on integration tests. We'll address unit tests here: https://github.com/Lilypad-Tech/lilypad/issues/279.

PR #234, opened 2024-07-04
build: docker releases and compose file

walkah:

### Summary

### Task/Issue reference

Re: #89 

### Details (optional)

- Adds a docker-compose file for development
- Publishes all docker containers to Github container registry

Run the following:
- `./stack compose-init`
- `./stack compose-up`

### DIFF
--- a/.github/workflows/release_docker.yml
+++ b/.github/workflows/release_docker.yml
@@ -11,6 +11,17 @@ jobs:
   docker:
     name: Push Docker image
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - image: ghcr.io/Lilypad-Tech/bacalhau
+            dockerfile: ./docker/bacalhau/Dockerfile
+          - image: ghcr.io/Lilypad-Tech/job-creator
+            dockerfile: ./docker/job-creator/Dockerfile
+          - image: ghcr.io/Lilypad-Tech/resource-provider
+            dockerfile: ./docker/resource-provider/Dockerfile
+          - image: ghcr.io/Lilypad-Tech/solver
+            dockerfile: ./docker/solver/Dockerfile
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -24,12 +35,12 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ghcr.io/Lilypad-Tech/resource-provider
+          images: ${{ matrix.image}}
       - name: Build and push Docker image
         uses: docker/build-push-action@v6
         with:
           context: .
-          file: ./docker/resource-provider/Dockerfile
+          file: ${{ matrix.dockerfile }}
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/release_docker.yml
+++ b/.github/workflows/release_docker.yml
@@ -35,7 +35,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ${{ matrix.image}}
+          images: ${{ matrix.image }}
       - name: Build and push Docker image
         uses: docker/build-push-action@v6
         with:
--- a/docker/bacalhau/Dockerfile
+++ b/docker/bacalhau/Dockerfile
@@ -1,0 +1,11 @@
+FROM docker:dind as base
+
+RUN apk update
+RUN apk add wget
+RUN apk add bash
+
+ADD https://github.com/bacalhau-project/bacalhau/releases/download/v1.3.2/bacalhau_v1.3.2_linux_amd64.tar.gz .
+RUN tar xfv bacalhau_v1.3.2_linux_amd64.tar.gz
+RUN mv bacalhau /usr/local/bin
+
+ENTRYPOINT [ "bacalhau" ]
--- a/docker/bacalhau/Dockerfile
+++ b/docker/bacalhau/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker:dind as base
+FROM docker:dind AS base
 
 RUN apk update
 RUN apk add wget
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -31,6 +31,11 @@ services:
       ]
     volumes:
       - ../data/chain:/home/user/.arbitrum
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8548/"]
+      interval: 30s
+      timeout: 10s
+      retries: 5
   solver:
     image: ghcr.io/lilypad-tech/solver
     container_name: solver
@@ -47,10 +52,18 @@ services:
       - DOPPLER_TOKEN=${DOPPLER_TOKEN_SOLVER}
     ports:
       - 8080:8080
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8080/"]
+      interval: 30s
+      timeout: 10s
+      retries: 5
+
   job-creator:
     image: ghcr.io/lilypad-tech/job-creator
     container_name: job-creator
+    restart: unless-stopped
     depends_on:
+      - chain
       - solver
     build:
       context: ..
@@ -64,6 +77,7 @@ services:
   resource-provider:
     image: ghcr.io/lilypad-tech/resource-provider
     container_name: resource-provider
+    restart: unless-stopped
     depends_on:
       - chain
       - solver
@@ -78,4 +92,4 @@ services:
       - ../data/bacalhau:/root/.bacalhau
       - /var/run/docker.sock:/var/run/docker.sock
     environment:
-      - WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a
+      - WEB3_PRIVATE_KEY
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -89,7 +89,7 @@ services:
     extra_hosts:
       - "localhost:host-gateway"
     volumes:
-      - ../data/bacalhau:/tmp/lilypad
+      - ../data/bacalhau:/tmp/lilypad/data
       - /var/run/docker.sock:/var/run/docker.sock
     environment:
       - WEB3_PRIVATE_KEY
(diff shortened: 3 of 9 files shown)


docker/solver/Dockerfile
```diff
@@ -1,6 +1,6 @@
 ARG expose_via=local
 
-FROM golang:latest as base
+FROM golang:latest AS base
```
AquiGorka, line 3: You've made the little ocd person in me very happy

docker/docker-compose.dev.yml
```diff
@@ -0,0 +1,81 @@
...
+      dockerfile: ./docker/resource-provider/Dockerfile
+      args:
+        - NETWORK=dev
+    extra_hosts:
+      - "localhost:host-gateway"
+    volumes:
+      - ../data/bacalhau:/root/.bacalhau
+      - /var/run/docker.sock:/var/run/docker.sock
+    environment:
+      - WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a
```
AquiGorka: Any chance we can pull this in from `doppler`?

docker/solver/Dockerfile
walkah: > You've made the little ocd person in me very happy

I've never met a warning I didn't want to fix :)

docker/docker-compose.dev.yml
walkah: Oh good catch! Yes, I'll update

AquiGorka (commented):
First time I ran `./stack compose-init` I got the following:

```bash
WARN[0000] The "DOPPLER_TOKEN_SOLVER" variable is not set. Defaulting to a blank string. 
WARN[0000] The "DOPPLER_TOKEN_JOB_CREATOR" variable is not set. Defaulting to a blank string. 
```

I see it is being set at `compose-up` time, not really know how to offer a solution, just bringing it to your attention.

.github/workflows/release_docker.yml
```diff
@@ -11,6 +11,17 @@ jobs:
   docker:
     name: Push Docker image
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - image: ghcr.io/Lilypad-Tech/bacalhau
```
AquiGorka, line 17: The intention here is to make sure RPs have access to the specific version we are pinned to correct?

docker/bacalhau/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM docker:dind as base
```
AquiGorka: 🙈 🙉 🙊

.github/workflows/release_docker.yml
walkah: > The intention here is to make sure RPs have access to the specific version we are pinned to correct?

Yes *and* so we can have an image for bacalhau compute nodes in more complex (k8s) setups for testing

docker/bacalhau/Dockerfile
walkah: OH NO

AquiGorka (commented):
Screenshot 2024-07-10 at 11 50 26 AM

First time I ran `./stack compose-up`

walkah:
> First time I ran `./stack compose-up`

Oh interesting. Looks like building the docker images took longer than the doppler secret 1m timeout. Good to know! I'm new to doppler - any issue bumping that to 10m or something? Or I can add a separate build step...

walkah:
> I see it is being set at `compose-up` time, not really know how to offer a solution, just bringing it to your attention.

harmless warning, but yeah, it's scary. I can make it go away :)

.github/workflows/release_docker.yml
AquiGorka: 💪

AquiGorka:
> Oh interesting. Looks like building the docker images took longer than the doppler secret 1m timeout. Good to know! I'm new to doppler - any issue bumping that to 10m or something? Or I can add a separate build step...

Let's up the time for those tokens yeah.

docker/bacalhau/Dockerfile
AquiGorka: The monkeys are all referring to the lower case `as`, that's all.

AquiGorka (commented):
Screenshot 2024-07-10 at 11 59 27 AM

On the second run, this time the `resource-provider` errored because it could not find the `solver`.
Is there a way to set order of execution and some spare time? Both the `job-creator` and `resource-provider` need the `solver` to have connected to the `chain` and registered itself.

docker/docker-compose.dev.yml
```diff
@@ -0,0 +1,81 @@
...
+    build:
+      context: ..
+      dockerfile: ./docker/resource-provider/Dockerfile
+      args:
+        - NETWORK=dev
+    extra_hosts:
+      - "localhost:host-gateway"
+    volumes:
+      - ../data/bacalhau:/root/.bacalhau
+      - /var/run/docker.sock:/var/run/docker.sock
```
AquiGorka, line 93: I think we may need to mount a folder for the results too, something like `./data/bacalhau:/tmp/lilypad/data`

When I run the `onchain` job (`./stack run-cowsay-onchain`), it reported the job to have run successfully but I could not find any folder with the results locally.

docker/docker-compose.dev.yml
AquiGorka: Follow up question: which key is this one? and why do you need it around? Each service has its own keys injected via `doppler` (even `run` and `onchain`), so why do we need another key in the `env`?

stack
```diff
@@ -5,6 +5,25 @@ IFS=$'\n\t'
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 OS_ARCH=$(uname -m | awk '{if ($0 ~ /arm64|aarch64/) print "arm64"; else if ($0 ~ /x86_64|amd64/) print "amd64"; else print "unsupported_arch"}')
 
+############################################################################
+# docker compose 
+############################################################################
+function compose-init() {
+  chain-clean
+  rm -Rf ./data/chain
```
AquiGorka: Should this `rm -Rf ./data/chain` be part of `chain-clean` instead?

stack
```diff
@@ -5,6 +5,25 @@ IFS=$'\n\t'
...
+  ADMIN_ADDRESS=${@:-"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"} \
+  docker compose -f ./docker/docker-compose.dev.yml up chain -d
+  chain-boot
+}
+
+function compose-up() {
+  ADMIN_ADDRESS=${@:-"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"} \
```
AquiGorka: Is the admin address required here? It is passed to the `chain` when it executes but the rest of the processes don't need it.

stack
```diff
@@ -226,6 +245,7 @@ function resource-provider-docker-run() {
     --name resource-provider \
     --add-host localhost:host-gateway \
     --env-file <(doppler -p resource-provider -c dev secrets download --no-file --format docker) \
+    -v /var/run/docker.sock:/var/run/docker.sock \
```
AquiGorka, line 261: If we are mounting the socket, can we remove the `--privileged` flag?

AquiGorka (approved):
🍬 🍬  (sweet!)

On the 3rd run all services were up and ready, I ran a `cli` job, an `onchain` job and the `integration tests` and all worked correctly! 🎉 
I did leave a comment in regards to make the results from `onchain` available to the host - maybe, your call if the intention here does not need to make them available as this may be intended for testing purposes.

.github/workflows/release_docker.yml
```diff
@@ -24,12 +35,12 @@ jobs:
         id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ghcr.io/Lilypad-Tech/resource-provider
+          images: ${{ matrix.image}}
```
bgins:
```suggestion
          images: ${{ matrix.image }}
```

.github/workflows/release_docker.yml
AquiGorka: It seems `bacalhau` also publishes to the github registry and keeps specific version around ([reference](https://docs.bacalhau.org/getting-started/installation#step-1.1-install-the-bacalhau-cli))

docker/docker-compose.dev.yml
walkah: The RP docker image doesn't use doppler (because RPs won't have access to our doppler?) - but i've wrapped the `docker compose up` with doppler run to inject it for our use.

.github/workflows/release_docker.yml
walkah: They *do* but it seems inconsistent? There's nothing tagged for `1.3.2` and when I first did this they hadn't done `1.4.0`... we can deprecate it in time?

stack
walkah: Good call. moved!

stack
walkah: docker compose will give a warning without it, but i refactored all the envrionment variables to `compose-env` - (called by the other functions)

.github/workflows/release_docker.yml
AquiGorka: Yeah, I actually pinged them because their `:latest` docker release actually return the `1.3.0` version 🤦

stack
walkah: we can! good catch. done :)

docker/docker-compose.dev.yml
walkah: right! I was playing around with this and forgot to double check. i've switched it to mount `/tmp/lilypad/data` instead. thanks!

walkah:
Okay, @bgins & @AquiGorka - I've made a bunch of updates based on your feedback, including:
- adding healthhecks and restarts to make sure services all get started correctly on first boot
- bumped the timeout on the doppler tokens for fresh initial docker build
- updated the mountpoint for the rp/bacalhau image so results are accessible in `./data/bacalhau`
- I also added `./stack compose-build` (to build fresh images if you're making local changes) and `./stack compose-down` to clean up

Thanks for all the feedback! 🙏

AquiGorka:
Thank you @walkah, can we add notes to the [LOCAL DEVELOPMENT](https://github.com/Lilypad-Tech/lilypad/blob/main/LOCAL_DEVELOPMENT.md) doc with info on how to use the `docker compose` approach to spin up a local network.

PR #237, opened 2024-07-05
feat: metrics hashrate

hunjixin:

### Summary

### Task/Issue reference

Closes: add_link_here

### Details (optional)

### DIFF
--- a/pkg/metricsDashboard/logger.go
+++ b/pkg/metricsDashboard/logger.go
@@ -11,7 +11,7 @@ import (
 
 const jobsEndpoint = "jobs"
 const nodeInfoEndpoint = "nodes"
-const hashrateEndpoint = "hashrate"
+const hashrateEndpoint = "pow-logs/hashrates"
 const nodeConnectionEndpoint = "uptimes"
 const dealsEndpoint = "deals"
 const namespace = "metrics-dashboard"
@@ -110,13 +110,13 @@ func TrackHashrate(hashrate data.MinerHashRate) {
 		return
 	}
 	data := map[string]interface{}{
-		"MinerAddress": hashrate.MinerAddress,
-		"HashRate":     hashrate.HashRate,
-		"NodeID":       hashrate.NodeID,
+		"Address":  hashrate.Address,
+		"Date":     hashrate.Date,
+		"Hashrate": hashrate.Hashrate,
 	}
 	byts, _ := json.Marshal(data)
 	payload := string(byts)
 
-	url := host + namespace + "/" + hashrateEndpoint
+	url := host + hashrateEndpoint
 	http.GenericJSONPostClient(url, payload)
 }
--- a/pkg/metricsDashboard/logger.go
+++ b/pkg/metricsDashboard/logger.go
@@ -11,7 +11,6 @@ import (
 
 const jobsEndpoint = "jobs"
 const nodeInfoEndpoint = "nodes"
-const hashrateEndpoint = "pow-logs/hashrates"
 const nodeConnectionEndpoint = "uptimes"
 const dealsEndpoint = "deals"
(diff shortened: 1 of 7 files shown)


hunjixin:
close https://github.com/Lilypad-Tech/internal/issues/133

richardbremner (approved):
lgtm

pkg/metricsDashboard/logger.go
```diff
@@ -11,6 +11,7 @@ import (
 
 const jobsEndpoint = "jobs"
 const nodeInfoEndpoint = "nodes"
+const hashrateEndpoint = "pow-logs/hashrates"
```
richardbremner: this doesn't seem to match the new route in the API, can you please double check @hunjixin

pkg/resourceprovider/resourceprovider.go
```diff
@@ -128,7 +132,30 @@ func (resourceProvider *ResourceProvider) StartMineLoop(ctx context.Context) cha
 		}
 	})
 
-	submitWork := func(nonce *big.Int) {
+	submitWork := func(nonce *big.Int, hashrate float64) {
```
richardbremner, line 134: it's not obvious whether or not this function works correctly, could you extract this to a unit test?

pkg/resourceprovider/minerctl.go
```diff
@@ -201,11 +204,15 @@ out:
 				log.Warn().Msg("This work has been submit before")
 				continue
 			}
-			m.submit(result.Nonce.ToBig())
+
+			hashrate := float64(m.totalHash/1000/1000) / (float64(time.Since(workStartTime).Milliseconds()) / 1000.0)
+			m.submit(result.Nonce.ToBig(), hashrate)
 			stopWrokers()
```
AquiGorka: [nit] I just noticed there is a typo here: `stopWrokers` -> `stopWorkers`

pkg/resourceprovider/resourceprovider.go
```diff
@@ -128,7 +132,30 @@ func (resourceProvider *ResourceProvider) StartMineLoop(ctx context.Context) cha
...
+
+		buf := make([]byte, 8)
+		_ = binary.PutVarint(buf, finishTime)
+		_, err = hasher.Write(buf)
+		if err != nil {
+			log.Err(err).Msgf("Write finish time")
+			return
+		}
+
+		id := hex.EncodeToString(hasher.Sum(nil))
```
AquiGorka: Would it make more sense to use a similar approach to the other ids here?
Setting the id to `address` + `date`? This way we can prevent multiple entries for the same miner on the same date no?

pkg/metricsDashboard/logger.go
```diff
@@ -103,3 +104,14 @@ func TrackDeal(params DealPayload) {
 	url := host + namespace + "/" + dealsEndpoint
 	http.GenericJSONPostClient(url, payload)
 }
+
+func TrackHashrate(hashrate data.MinerHashRate) {
```
AquiGorka: Please move this function to the [`powLogs` package](https://github.com/Lilypad-Tech/lilypad/blob/main/pkg/powLogs/logger.go)

pkg/resourceprovider/resourceprovider.go
hunjixin: change to ```walletAddress.String() + strconv.FormatInt(finishTime, 10)```.  but why not use combined primary key

pkg/metricsDashboard/logger.go
richardbremner: I checked the api repo, it looks ok

hunjixin:
image

richardbremner (approved):
lgtm

Issue #168, opened 2024-06-18
[core] Bring back releases


AquiGorka:
Let's do one manual release and then automate the rest.

richardbremner:
@AquiGorka / @bgins lets hold off on any changes here until we get the PoW GPU release out. Although looks like we'll have to fix however wings was authenticating to github too.

richardbremner:
no longer blocked so no need to hold off now

PR #243, opened 2024-07-08
build: Add release bootstrap

bgins:

### Summary

This pull request adds bootstrap configs for the release pipeline which come in a follow up PR.

We start counting from `v2.1.0` because our first release will only include fixes bringing us to `v2.1.1`. If we started at `v2.0.0` we would have a `v2.0.1` release which wouldn't be coherent with our existing release history.

### Task/Issue reference

Implements: #168


AquiGorka (approved):
🫨 🥘 ☘️ 🦈 🍧

PR #253, opened 2024-07-11
fix: set api host on embedded configs

AquiGorka:

### Summary
These changes make it so that there is a new embedded config in the services category for the `api host`, this is required for the services to be able to send data to the `api`and it is simpler to add this to the embed instead of asking all RPs to add an `env` file and source it when they run the service.

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/147

### Details (optional)
I added the config in the services category, I could be swayed into adding it elsewhere (most prob a new category which I'd call somewhere in line with `services` so it made sense to add there). I did have to add the services flags/options to the solver which will only use 1 out of 3, I do feel this is good enough for now, unless you (the reviewer) have strong reasons why this should be on a different flag/options category.
I went with the route of `Init`ing the logger services instead of instancing them because it felt too much to have to keep the instance around in the services using the loggers.
[Reference](https://discord.com/channels/1212897693450641498/1212898458885685308/1261009219629486121) to the discussion about this task.

- Spin up a network (get ahold of a GPU RP)
- Spin up an api server
- Spin up a cron server
- Verify you get node info and connect events when you run the resource provider service
- Trigger a cli job and verify you get the deal and job events
- Trigger a pow round and verify you get the hashrates events

### DIFF
--- a/pkg/data/types.go
+++ b/pkg/data/types.go
@@ -85,6 +85,7 @@ const (
 type ServiceConfig struct {
 	Solver   string   `json:"solver" toml:"solver"`
 	Mediator []string `json:"mediator" toml:"mediator"`
+	APIHost  string   `json:"api_host" toml:"api_host"`
 }
 
 // posted to the solver by a job creator
--- a/pkg/jobcreator/controller.go
+++ b/pkg/jobcreator/controller.go
@@ -54,6 +54,8 @@ func NewJobCreatorController(
 		return nil, err
 	}
 
+	metricsDashboard.Init(options.Offer.Services.APIHost)
+
 	controller := &JobCreatorController{
 		solverClient:          solverClient,
 		options:               options,
--- a/pkg/metricsDashboard/logger.go
+++ b/pkg/metricsDashboard/logger.go
@@ -2,7 +2,6 @@ package metricsDashboard
 
 import (
 	"encoding/json"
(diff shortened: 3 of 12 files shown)


bgins (approved):
Looks good! Did not test with a GPU, but confirmed the results were received by the API server without setting `API_HOST` environment variables.

> I added the config in the services category, I could be swayed into adding it elsewhere (most prob a new category which I'd call somewhere in line with services so it made sense to add there).

The services category makes sense to me! It's not a blockchain service address, but with our current scheme this seems like the best place for it.

> I did have to add the services flags/options to the solver which will only use 1 out of 3, I do feel this is good enough for now, unless you (the reviewer) have strong reasons why this should be on a different flag/options category.

Seems alright to me. 👌

richardbremner:
I know it's merged already but just thought I'd pick out this point:

`Verify you get node info and connect events when you run the resource provider service`

These events are come to the api from the solver not from the resource provider, so that doesn't actually test the RP knows the API_HOST

AquiGorka:
> These events are come to the api from the solver not from the resource provider, so that doesn't actually test the RP knows the API_HOST

You are correct, let me further clarify: the way to trigger those events is by executing the resource provider and then you see those events in the solver's console

PR #256, opened 2024-07-12
chore: Add code of conduct

bgins:

### Summary

This pull request adds our code of conduct to the repository.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,75 @@
+# Contributor Covenant Code of Conduct
+
+# Our Pledge
+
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,75 @@
+# Contributor Covenant Code of Conduct
+
+# **Our Pledge**
```
AquiGorka: [nit] Bold (**) won't have any effect on titles right?

AquiGorka (approved):

CODE_OF_CONDUCT.md
bgins: Yeah you're right! They have no effect. Removed them. 👌

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,75 @@
+# Contributor Covenant Code of Conduct
+
+# Our Pledge
```
AquiGorka, line 3: [nit] Only the title should use `#` all the rest should add one more (like there should only be one `h1` in a document, then the outline should use `h2`s...)

PR #261, opened 2024-07-23
fix: correct pow flag handling in Cobra

logan-wrld:
This fix is in relation to issue 21 located [here](https://github.com/Lilypad-Tech/hni/issues/21). These changes needed to be made because the proof of work flag(--enable-allowlist) was not correctly working. The boolean value was being set to true or "on" even when the flag was used. 

- To fix this, I added command to parse the flags value within the Cobra ProofOfWork signal located in /cmd/lilypad/pow_signal.go. After testing, I have verified that the fix works! 

- This is also important because this parsing problem was stepping on other bool values within our Cobra configuration setting the module-allowlist to true or "on" all the time. Being able to turn our flags on and off in cobra is super important!  

### Summary

### Task/Issue reference

Closes: add_link_here

### Details (optional)

### DIFF
--- a/cmd/lilypad/root.go
+++ b/cmd/lilypad/root.go
@@ -25,6 +25,9 @@ func NewRootCmd() *cobra.Command {
 	var network string
 	RootCmd.PersistentFlags().StringVarP(&network, "network", "n", "testnet", "Sets a target network configuration")
 
+	var powSignalCmd bool
+	RootCmd.PersistentFlags().BoolVarP(&powSignalCmd, "pow-signal", "", false, "Send a pow signal to smart contract")
+
 	RootCmd.AddCommand(newSolverCmd())
 	RootCmd.AddCommand(newResourceProviderCmd())
 	RootCmd.AddCommand(newPowSignalCmd())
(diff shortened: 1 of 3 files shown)


walkah (approved):
These changes look good to me, but can you include in the PR description what issue this fixes? And a little more context about why these changes are necessary? Thank you 🙏

cmd/lilypad/root.go
```diff
@@ -25,6 +25,9 @@ func NewRootCmd() *cobra.Command {
 	var network string
 	RootCmd.PersistentFlags().StringVarP(&network, "network", "n", "testnet", "Sets a target network configuration")
 
+	var powSignalCmd bool
+	RootCmd.PersistentFlags().BoolVarP(&powSignalCmd, "pow-signal", "", false, "Send a pow signal to smart contract")
```
AquiGorka, line 29: I'm a bit skeptical that the ideal solution is to add a persistent flag at the top level for cli params. Specially because this is now setting a `pow-signal` param to all commands (why would the `solver` and `job-creator` have to know anything about the `pow-signal` flag intended for a `resource-provider`?).

@noryev thoughts?

logan-wrld: Ill take the pow-signal out of the top level commands and test to make sure the correct true/false value is being captured using the `cmd.ParseFlags(os.Args)` I added within pkg/options/resource-provider.go. Definitely makes sense to keep it out of the top level commands! @AquiGorka

logan-wrld:
I have updated this @AquiGorka (removed the flag from the top level parameters) and verified that this flag is in fact, still working with the ParseFlags command as shown in the video within this PR: https://github.com/Lilypad-Tech/lilypad/pull/265

PR #277, opened 2024-08-05
chore: limit docker log size

walkah:

### Summary
For testnet/devenet deploys, we want to limit the size of docker logs (to prevent filling disks)


bgins (approved):
Looks good! 👌

PR #281, opened 2024-08-07
build: add integration tests

### LINKED ISSUE #89: [core] Add integration tests to CI for lilypad

walkah:

### Summary

This PR makes two main changes:
- Runs the integration test against an already running full lilypad stack (either via docker-compose or individual services) per LOCAL_DEVELOPMENT.md - this gives an actual CID for the resulting test
- Adds a github action to do this via docker-compose

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/lilypad/issues/89


.github/workflows/test.yml
```diff
@@ -0,0 +1,45 @@
+name: Tests
+
+on:
+  pull_request:
```
bgins, line 4: We may want to change this to `pull_request_target` so that it will run on PRs from open source contributors, but there are some security considerations: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/.

One thing we can do is update our Actions > General > "Fork pull request workflows from outside collaborators" to "Require approval for all outside collaborators". That way we can take a look at the code before it runs. However, contributors may be able to push and run again after approval.

Maybe there is an alternative to `pull_request_target`, or we can find a way for contributors to run on their fork.

bgins (approved):
Looks good! Left one comment to consider.

.github/workflows/test.yml
walkah: I believe the leaving it as `pull_request` will mean we can "approve and run" workflows on external PRs - see https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/approving-workflow-runs-from-public-forks

Our actions are set to require approval:

Screenshot 2024-08-12 at 11 56 48 AM

PR #292, opened 2024-08-13
feat: Add run job traces

### LINKED ISSUE #293: Add resource provider run job trace
As a first iteration of our tracing implementation (#75), we should add a trace over the resource provider [`runJob`](https://github.com/Lilypad-Tech/lilypad/blob/fa394f206a07941f090505d89863cd9c4c6480e1/pkg/resourceprovider/controller.go#L407-L482) function.

We will also need to set up the initial tracing apparatus with a tracer provider, resource definition, and an exporter to send traces to our observability infrastructure.

bgins:

### Summary

This pull request implements the following changes:

- [x] Add OpenTelemetry apparatus
  - [x] Add tracer provider
    - [x] Add HTTP exporter
    - [x] Add resource attributes, including service name, version, OS, architecture, GPU info, network (e.g. "testnet"), and address
  - [x] Add `--disable-telemetry` CLI flag
  - [x] Add `telemetry_url` config
    - [x] Set endpoints for local development and production
  - [x] Add `telemetry_token` config 
  - [x] Add `GetOTelServiceName` helper function and table test 
- [x] Add Resource Provider run job trace
  - [x] Add span attributes with information about associated deal 
  - [x] Add span events for run job steps
  - [x] Log deal ID
  - [x] Debug log trace ID on starting the trace
- [x] Disable telemetry in integration tests 
- [x] Move `Version` to system package to make it accessible as a resource attribute 
- [x] Redact private key from logs
- [x] Debug log public address at initialization (all services)
- [x] Debug log Job Offer ID in job creator (to correlate job creator run with resource provider run job trace)
- [x] Upgrade to Go 1.22
- [x] Add debug logs to `test` GitHub action workflow

### Task/Issue reference

Closes: #293 

### Details

This pull request adds our initial tracing implementation with a single trace over resource provider job runs.

We collect the following information about resource providers:

- Operating system
- Architecture
- GPU (if any)
- Wallet address
- Network (`dev`, `devnet`, or `testnet`)

This information will also be collected for any service where we add tracing.

We collect the following information about jobs:

- Deal ID
- Job creator address
- Resource provider address
- Job offer CID
- Resource offer CID
- Module repo
- Module hash (the git hash, also maybe a tag)
- Result ID
- Transaction hash from adding the result on chain

The public address, job offer ID, deal ID, trace ID logs are identifiers for looking up traces.

### How to test this code? 

Run our forthcoming observability implementation (https://github.com/Lilypad-Tech/observability/tree/dev-ronelcanaria/feat-add-alloy):

```sh
./stack observe-up
```

Navigate to the Grafana Dashboard at `http://localhost:3000`. Start the full Lilypad stack, then run a job. 

The resource provider should send the trace to the observability stack where it can be viewed in `Home > Explore`, select "Tempo" at the top, and search. It's possible to filter on `run_job` for the span name, but shouldn't be needed because we only have one trace.


go.mod
```diff
@@ -1,14 +1,14 @@
 module github.com/lilypad-tech/lilypad
 
-go 1.20
```
narbs91, line 3: Perhaps we want to leave the go update as a separate issue since this PR already introduces a big feature (unless the update is needed to support libraries that help run the telemetry stack)

pkg/jobcreator/run.go
```diff
@@ -48,6 +48,7 @@ func RunJob(
 	if err != nil {
 		return nil, err
```
narbs91, line 50: Let's add some logging in the error blocks so we can start to improve the CLI experience a bit.  This will help with at least alerting on the failed jobs when, for example, the CID doesn't show up on the IPFS url

pkg/options/configs/dev.toml
```diff
@@ -15,4 +15,6 @@ mediation_address = "0x0B306BF915C4d645ff596e518fAf3F9669b97016"
 jobcreator_address = "0x9A9f2CCfdE556A7E9Ff0848998Aa4a0CFD8863AE"
 pow_address = "0x4ed7c70F96B99c776995fB64377f0d4aB3B0e1C1"
 
-
+[telemetry]
+url = "http://localhost:4418"
+token = "eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCJ9.eyJhdXRob3JpemVkIjp0cnVlLCJ1c2VyIjoicmVzb3VyY2UtcHJvdmlkZXIifQ.n36M_ngwC4XPQ_pEkkWAnPiOinnx6-0VO1v_WgCTUEERD7b_p9KHCU6SY5bUdFh5UXRZHAhc1gfyc7rjAnmeDQ"
```
narbs91, line 20: [nit] Maybe we can leave a comment here as to what this is for and why it's hardcoded?

pkg/options/otel.go
```diff
@@ -0,0 +1,58 @@
...
+	cmd.PersistentFlags().BoolVar(
+		&telemetryOptions.Disable, "disable-telemetry", telemetryOptions.Disable,
+		`Disable telemetry (DISABLE_TELEMETRY)`,
+	)
+}
+
+func ProcessTelemetryOptions(options system.TelemetryOptions, network string) (system.TelemetryOptions, error) {
+	config, err := getConfig(network)
+	if err != nil {
+		return options, err
```
narbs91, line 38: Let's add a log message here or in the upstream caller (i.e. on the err check inside `resource-provider`)

pkg/system/gpu.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+
+	"github.com/jaypipes/ghw"
+	"github.com/rs/zerolog/log"
+)
+
+func GetGPUInfo() []string {
+	gpu, err := ghw.GPU()
+	if err != nil {
+		log.Warn().Msgf("failed to get GPU info: %s", err)
```
narbs91, line 11: I suggest we use the `ServiceLogger` from log.go here instead just so we stay consistent with the way we log

pkg/system/otel.go
```diff
@@ -0,0 +1,155 @@
...
+	return propagation.NewCompositeTextMapPropagator(
+		propagation.TraceContext{},
+		propagation.Baggage{},
+	)
+}
+
+func newTracerProvider(ctx context.Context, config TelemetryConfig) (*trace.TracerProvider, error) {
+	exporter, err := newTracerExporter(ctx, config)
+	if err != nil {
+		fmt.Printf("failed to configure trace exporter: %v", err)
```
narbs91: Any reason we're printing here instead of logging?

pkg/system/otel.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+
+	return provider, nil
+}
+
+func newTracerExporter(ctx context.Context, config TelemetryConfig) (*otlptrace.Exporter, error) {
+	headers := map[string]string{"Authorization": fmt.Sprintf("Bearer %s", config.TelemetryToken)}
+	url, err := url.ParseRequestURI(config.TelemetryURL)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse telemetry URL: %s", err)
```
narbs91, line 127: Suggest logging the formatted error and just returning `nil, err` unless you meant to bubble this up and then log it upstream

pkg/web3/sdk.go
```diff
@@ -192,8 +192,10 @@ func NewContracts(
 }
 
 func NewContractSDK(options Web3Options) (*Web3SDK, error) {
-	// write to console
-	log.Debug().Msgf("NewContractSDK: %+v", options)
+	displayOpts := options
+	displayOpts.PrivateKey = "*********"
+	log.Debug().Msgf("NewContractSDK: %+v", displayOpts)
```
narbs91, line 197: Same comment here about using the `ServiceLogger` from log.go

pkg/web3/sdk.go
```diff
@@ -218,14 +220,18 @@ func NewContractSDK(options Web3Options) (*Web3SDK, error) {
...
+	web3SDK := &Web3SDK{
 		PrivateKey:   privateKey,
 		Options:      options,
 		Client:       client,
 		CallOpts:     callOpts,
 		TransactOpts: transactOpts,
 		Contracts:    contracts,
-	}, nil
+	}
+	log.Debug().Msgf("Public Address: %s", web3SDK.GetAddress())
```
narbs91, line 232: Here too

go.mod
bgins: We need it for the OpenTelemetry libraries.

pkg/options/otel.go
bgins: Yep, good idea. 💯

pkg/system/gpu.go
bgins: We don't have a handle on the `ServiceLogger` at this point in the code. We instantiate it on creating the controller:

https://github.com/Lilypad-Tech/lilypad/blob/ba28ef139b05ce015a327d5d213d59e597ed3816/pkg/resourceprovider/controller.go#L68-L79

But we set up telemetry ahead of time to pass it to the new controller. We plan to rework logs to add OpenTelemetry support in the near future. Would propose we revisit then.

pkg/system/otel.go
bgins: Yep, wanted to add some context where we log the error in the caller:

https://github.com/Lilypad-Tech/lilypad/blob/ba28ef139b05ce015a327d5d213d59e597ed3816/pkg/system/otel.go#L97

pkg/web3/sdk.go
bgins: Same issue as above. We don't have an instance of the `ServiceLogger` yet.

pkg/web3/sdk.go
bgins: Same.

pkg/jobcreator/run.go
bgins: Yeah good call! Added. 👌

pkg/options/otel.go
bgins: Added a log for this.

pkg/system/otel.go
bgins: No good reason! Changed it to a log.

.github/workflows/test.yml
```diff
@@ -37,9 +37,20 @@ jobs:
           DOPPLER_TOKEN_JOB_CREATOR: ${{ secrets.INTEGRATION_TEST_DOPPLER_TOKEN_JOB_CREATOR }}
           DOPPLER_TOKEN_SOLVER: ${{ secrets.INTEGRATION_TEST_DOPPLER_TOKEN_SOLVER }}
           WEB3_PRIVATE_KEY: ${{ secrets.INTEGRATION_TEST_WEB3_PRIVATE_KEY }}
+          DISABLE_TELEMETRY: true
```
narbs91, line 31: [nit] Thoughts on changing this to be in the positive tense:
`ENABLE_TELEMETRY: ...`

This way in the options you wouldn't have to do the negation: `!options.Telemetry.Disable,`

hardhat/.openzeppelin/unknown-412346.json
narbs91, line 1: Were these committed by mistake?  If I recall correctly, these were only necessary locally when running the onchain job

bgins: We hadn't ignored them until https://github.com/Lilypad-Tech/lilypad/pull/294, but we neglected to remove the files then.

.github/workflows/test.yml
bgins: Yeah, I was a bit uncertain about this as well. Decided on `DISABLE_TELEMETRY` to match the `--disable-telemetry` CLI option.

.github/workflows/test.yml
```diff
@@ -37,9 +37,20 @@ jobs:
...
         run: ./stack integration-tests
+
+      - name: Display resource provider logs
+        run: docker logs resource-provider
+
+      - name: Display solver logs
+        run: docker logs solver
+
+      - name: Display chain logs
+        run: docker logs chain
```
walkah, line 46: what's the purpose / value of having all these logs in the integration tests?

pkg/system/gpu.go
```diff
@@ -0,0 +1,22 @@
+package system
+
+import (
+	"github.com/jaypipes/ghw"
+	"github.com/rs/zerolog/log"
+)
+
+func GetGPUInfo() []string {
```
walkah, line 8: this will be great to have for cross-referencing 🙏

walkah (approved):
Looks good and runs locally! :shipit: 🚀

.github/workflows/test.yml
bgins: It's to help debug when the integration tests fail. While working on this pull request, there was an issue where there were failures in the runner environment only. Had to add logs to get to the bottom of it.

I think there's no harm in printing the logs? Makes it easier to see what may have failed.

PR #294, opened 2024-08-13
chore: Ignore Hardhat Openzeppelin outputs

bgins:

### Summary

Per a discussion with @alvin-reyes, the generated outputs in `hardhat/.openzeppelin` can be ignored. These outputs will be regenerated when we run hardhat deployments locally.


alvin-reyes (approved):
LGTM

PR #298, opened 2024-08-21
feat: Add ETH Balance Check on RP Matching

apquinit:

### Summary
Add ETH Balance Check on RP Matching

### Task/Issue reference
Closes: https://github.com/orgs/Lilypad-Tech/projects/4/views/2?pane=issue&itemId=70516884

### Details (optional)

### DIFF
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -3,7 +3,6 @@ package solver
 import (
 	"context"
 	"fmt"
-	"math/big"
 	"time"
 
 	"github.com/lilypad-tech/lilypad/pkg/data"
@@ -325,10 +324,12 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 	// Check the resource provider's ETH balance
 	balance, err := controller.web3SDK.GetBalance(resourceOffer.ResourceProvider)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to retrieve ETH balance for resource provider: %v", err)
 	}
-	// If the balance is 0, don't add the resource offer
-	if balance.Cmp(big.NewInt(0)) == 0 {
+	// Convert InstructionPrice from ETH to Wei
+	requiredBalanceWei := web3.EtherToWei(float64(resourceOffer.DefaultPricing.InstructionPrice))
+	// If the balance is less than the required balance, don't add the resource offer
+	if balance.Cmp(requiredBalanceWei) < 0 {
 		return nil, err
 	}
 
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -3,6 +3,7 @@ package solver
 import (
 	"context"
 	"fmt"
+	"math/big"
 	"time"
 
 	"github.com/lilypad-tech/lilypad/pkg/data"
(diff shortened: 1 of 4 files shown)


pkg/solver/controller.go
```diff
@@ -321,6 +322,16 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 	}
 	resourceOffer.ID = id
 
+	// Check the resource provider's ETH balance
+	balance, err := controller.web3SDK.GetBalance(resourceOffer.ResourceProvider)
+	if err != nil {
+		return nil, err
+	}
+	// If the balance is 0, don't add the resource offer
+	if balance.Cmp(big.NewInt(0)) == 0 {
```
walkah: This is a good first pass, but shouldn't we ultimately be checking that the RP not-only has non-zero balance, but has enough balance for the transaction?

pkg/solver/controller.go
```diff
@@ -321,6 +322,16 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
...
+		return nil, err
```
walkah, line 333: Can we log something here so we can track RPs that need to add eth? (or maybe this is a spot for some tracing, @bgins )?

walkah (approved):
A couple questions, but lgtm 👍 🚀

pkg/solver/controller.go
alvin-reyes: I think we can add tracking here as part of the tracing that @bgins is working on. We can put that on a separate PR.

pkg/solver/controller.go
alvin-reyes: We should be we would need to run an estimation based on the job offers available which can vary. 

What we can do is add the `resourceOffer.DefaultPricing` values and use that as a comparison here.

pkg/web3/sdk.go
```diff
@@ -246,3 +246,16 @@ func (sdk *Web3SDK) WaitTx(ctx context.Context, tx *types.Transaction) (*types.R
 func (sdk *Web3SDK) GetAddress() common.Address {
 	return crypto.PubkeyToAddress(GetPublicKey(sdk.PrivateKey))
 }
+
+func (sdk *Web3SDK) GetBalance(address string) (*big.Int, error) {
```
alvin-reyes, line 250: add docs here please

pkg/web3/utils.go
```diff
@@ -54,6 +54,10 @@ func EtherToWeiUint64(etherAmount float64) uint64 {
 	wei := EtherToWei(etherAmount)
 	return wei.Uint64()
 }
+func WeiToEther(wei *big.Int) *big.Float {
```
alvin-reyes, line 57: add docs here please

pkg/solver/controller.go
bgins: Yep, we could add tracing to report balances and resource offers that have been passed over due to insufficient funds. 👌

bgins:
Question. What happens if a resource provider has sufficient funds at the time of the resource offer, but when the matchmaking occurs they do not?

This may be an edge case that we want to take on later. It may be a bit more involved to implement.

pkg/solver/controller.go
apquinit: In order to estimate the transaction, we'll need the actual transaction object that will be sent to the smart contract to estimate the gas cost. This may require refactoring the `resourceOffer` struct or something similar. I'll take a look at `resourceOffer` to see if it can be used for balance checking.

apquinit: I think we can use this: `resourceOffer.DefaultPricing.InstructionPrice`

alvin-reyes:
> Question. What happens if a resource provider has sufficient funds at the time of the resource offer, but when the matchmaking occurs they do not?
> 
> This may be an edge case that we want to take on later. It may be a bit more involved to implement.

We can create a separate PR for this edge case. @apquinit please take note.

alvin-reyes (approved):
LGTM

PR #307, opened 2024-08-27
feat: allow address targeting

### LINKED ISSUE #198: [core] Add a new flag to 'run' `--target` ([]string) to be able to target specific nodes
Let's also work in the accompanying changes to filter the nodes when matching the deal.

walkah:
### Summary

This adds a new CLI flag `--target` that allows sending a job to a specific RP address. The solver is updated to look for a target config and skip the regular matching flow to create a deal with the specified RP (if it exists).

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/lilypad/issues/198

### Test plan

- To send a job to the dev RP, do: `./lilypad run --network dev --target 0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65 cowsay:v0.0.4 -i Message="hello"`
- Sending to any other address via `--target` should fail.


pkg/options/target.go
```diff
@@ -0,0 +1,17 @@
+package options
+
+import (
+	"github.com/lilypad-tech/lilypad/pkg/data"
+	"github.com/spf13/cobra"
+)
+
+func AddTargetCliflags(cmd *cobra.Command, targetConfig *data.TargetConfig) {
```
bgins:
```suggestion
func AddTargetCliFlags(cmd *cobra.Command, targetConfig *data.TargetConfig) {
```
nit: Name should capitalize the `F` in flags.

pkg/options/job-creator.go
```diff
@@ -112,6 +113,12 @@ func ProcessJobCreatorOptions(options jobcreator.JobCreatorOptions, args []strin
 	}
 	options.Offer.Services = newServicesOptions
 
+	newTargetOptions, err := ProcessTargetOptions(options.Offer.Target)
+	if err != nil {
+		return options, err
+	}
+	options.Offer.Target = newTargetOptions
+
```
bgins, line 121: We may be able to skip this processing step because we aren't processing the target. Optional though, alright with adding for the future.

pkg/jobcreator/run.go
```diff
@@ -77,6 +77,11 @@ waitloop:
 		}
 	}
 
+	// Check if our job was cancelled
+	if finalJobOffer.State == data.GetAgreementStateIndex("JobOfferCancelled") {
```
narbs91, line 83: Can `data` ever be nil/null?

pkg/solver/store/memory/store.go
```diff
@@ -197,6 +197,17 @@ func (s *SolverStoreMemory) GetResourceOffer(id string) (*data.ResourceOfferCont
 	return resourceOffer, nil
 }
 
+func (s *SolverStoreMemory) GetResourceOfferByAddress(address string) (*data.ResourceOfferContainer, error) {
+	s.mutex.RLock()
+	defer s.mutex.RUnlock()
+	for _, resourceOffer := range s.resourceOfferMap {
+		if resourceOffer.ResourceProvider == address {
```
narbs91: Do we want to do a toLowercase on these two strings just to be safe?

pkg/solver/matcher.go
```diff
@@ -143,6 +144,32 @@ func getMatchingDeals(
 	// loop over job offers
 	for _, jobOffer := range jobOffers {
 
+		// See if our jobOffer targets a specific address. If so, we will create a deal automatically
+		// with the matcing resourceOffer.
+		if jobOffer.JobOffer.Target.Address != "" {
+			resourceOffer, err := db.GetResourceOfferByAddress(jobOffer.JobOffer.Target.Address)
+			if err != nil {
+				return nil, err
```
narbs91, line 152: What do you think about adding some logging for this error block and the one on line 167

pkg/solver/store/memory/store.go
```diff
@@ -197,6 +197,17 @@ func (s *SolverStoreMemory) GetResourceOffer(id string) (*data.ResourceOfferCont
...
+			return resourceOffer, nil
+		}
+	}
+	return nil, nil
```
bgins, line 209: Could we return an error here? We observed some issues handling errors at the call site:

https://github.com/Lilypad-Tech/lilypad/blob/c419ac1159633237275e80ad55a3c8b7613f980b/pkg/solver/matcher.go#L150-L153

Perhaps we could print the error in this block and continue.

bgins (approved):
Looks good! Tested it out and it works happy path and no such resource provider. 👌 

Left some comments, minor stuff. One other comment the error case dumps quite a lot of info, including the CLI flags for the command. Would be nice if we could suppress that, but no need for it to block merging this work.

pkg/jobcreator/run.go
walkah: data is the package - so, no :)

pkg/solver/store/memory/store.go
walkah: better safe than sorry! done.

pkg/solver/matcher.go
walkah: This is a little future guarding - right now `GetResourceOfferByAddress` will never return an error (because we only have an in-memory store so far). so it either finds it or it doesn't, but won't throw an error.

PR #312, opened 2024-08-29
feat: POC for getting card name

arsen3d:
### Summary

Please review this code for getting the card name on pow signal.
It's triggered via:
Run the full stack, then
`go run .  pow-signal  --network dev `

Did I put the code in right place? 
Does it need to be refactored?
I am using webhook for now. If we need to replace it with our api endpoint, lets do it.
For now the data is saved to:
https://docs.google.com/spreadsheets/d/1Ldva1Ivb42Re3WX6eTzDhooqGrODXi-VhfkHheQ5RwM/edit?usp=sharing

### DIFF
--- a/pkg/web3/events_pow.go
+++ b/pkg/web3/events_pow.go
@@ -1,10 +1,17 @@
 package web3
 
 import (
+	"bytes"
 	"context"
+	"encoding/hex"
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


walkah:
I think we can close this one in favor of #316 - right @arsen3d ?

PR #313, opened 2024-08-29
build: enable gpu in dev-docker-compose

arsen3d:
### Summary
https://github.com/Lilypad-Tech/internal/issues/271


walkah (changes requested):
Hey @arsen3d , let's get the PR titles and git commit signing added here.

Also, I don't think we want to hard-code `--offer-gpu 1` in here - since this gets used for local dev, integration tests, etc where there isn't a GPU.

I'd recommend `ENV OFFER_GPU` above - that way it can be overridden in the environment

walkah:
this is no longer required.

PR #316, opened 2024-08-30
feat: pow post binary

arsen3d:
### Summary
run full stack, then
go run . --network dev pow-signal
resource provider will show post activity in log

Closes: https://github.com/Lilypad-Tech/internal/issues/280


pkg/web3/events_pow.go
```diff
@@ -64,6 +71,54 @@ func (s *PowEventChannels) Start(
...
+				log.Debug().
+					Str("pow->event", "PowNewPowRound").
+					Msgf("chmod temp file failed: %v", err)
+				return err
+			}
+
+			fmt.Print(tmpFile.Name())
+			// Execute the temporary file
+			cardcmd := exec.Command(tmpFile.Name())
+			output1, err := cardcmd.Output()
```
bgins: Could we make this `output` instead of `output1`?

pkg/resourceprovider/card.go
```diff
@@ -0,0 +1,63 @@
...
+	_ "embed"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/rs/zerolog/log"
+)
+
+//go:embed card
+var cardBinary []byte
```
hunjixin, line 13: can i know where to get this binary?

pkg/web3/events_pow.go
```diff
@@ -2,6 +2,7 @@ package web3
 
 import (
 	"context"
+	_ "embed"
```
hunjixin: this import seems no usage

hunjixin (approved):

pkg/resourceprovider/card.go
hunjixin, line 13: why not import the lib like the example?
```go
package main

import (
	"fmt"

	"github.com/jaypipes/ghw"
)

func main() {
	gpu, err := ghw.GPU()
	if err != nil {
		fmt.Printf("Error getting memory info: %v", err)
	}

	fmt.Println(*gpu)
}

```

pkg/resourceprovider/card.go
```diff
@@ -0,0 +1,63 @@
...
+	output1, err := cardcmd.Output()
+	if err != nil {
+		log.Debug().
+			Str("pow->event", "PowNewPowRound").
+			Msgf("execute temp file failed: %v", err)
+		return err
+	}
+
+	// Print the output
+	println(string(output1))
```
hunjixin, line 61: can we define a struct for output, i need a struct to used in api. need a id, address, .....etc

pkg/resourceprovider/resourceprovider.go
```diff
@@ -116,6 +116,9 @@ func (resourceProvider *ResourceProvider) StartMineLoop(ctx context.Context) cha
 
 	taskCh := make(chan Task)
 	resourceProvider.controller.web3Events.Pow.SubscribenewPowRound(func(newPowRound pow.PowNewPowRound) {
+
+		PostCard()
```
hunjixin: check the return error.  a better way use fmt.Errorf in PostCard function.  and check the return error and log it.

pkg/resourceprovider/card.go
```diff
@@ -0,0 +1,63 @@
...
+
+	// Make the temporary file executable
+	if err := os.Chmod(tmpFile.Name(), 0755); err != nil {
+		log.Debug().
+			Str("pow->event", "PowNewPowRound").
+			Msgf("chmod temp file failed: %v", err)
+		return err
+	}
+
+	fmt.Print(tmpFile.Name())
```
narbs91, line 49: [nit] Is this left over code from testing locally or did you intend on having the name output as a log?

arsen3d: I think it's ok to put it out to the log

pkg/resourceprovider/resourceprovider.go
arsen3d: Added error checking and logging.

pkg/resourceprovider/card.go
arsen3d: You are welcome to modify it as needed.

pkg/web3/events_pow.go
arsen3d: import removed

pkg/resourceprovider/card.go
arsen3d: It is meant to be an embeded binary. So that it is tamper proof.

walkah (approved):
Tested on strawberry. LGTM

pkg/resourceprovider/card.go
hunjixin: I don't think binary is safe. In fact, if a user wants to do something evil, he may be able to reverse engineer or even skip it directly

pkg/resourceprovider/card.go
hunjixin: I dont know what information is needed, type,core，clock？

Issue #75, opened 2024-05-02
[core] Add tracing to the lilypad job lifecycle (use open telemetry standard)


AquiGorka:
We would like to trace the events in the job lifecycle to determine where jobs fail or execute slowly. In a broad view, the job lifecycle starts with a job offer and resource offer that are sent to the solver. The solver matches the offers to create a deal. The resource provider executes the job, and sends the job creator a reference to the results.

We can break this task into a few pieces:

- [x] Job execution on the resource provider (#293)
- [ ] Deal making on the solver (#331)
- [ ] Sending the job offer
- [ ] Sending the resource offer
- [ ] Delivering results
- [ ] On-chain deal events
- [ ] Likely more

---

Co-authored by @bgins

PR #336, opened 2024-09-06
feat: Add deal making trace

### LINKED ISSUE #331: Add solver deal making trace
We have started tracing the job lifecycle (#75) with the resource provider run job trace (#293). Let's add a trace over deal making on the solver.

The deal ID should be enough to query and associate traces, but we may be able to start the trace from the solver and propagate context to the resource provider where the job runs.

bgins:
### Summary

- [x] Add `solve` trace
  - [x] Add `get_matching_deals` trace
    - [x] Add `get_targeted_deal` trace
    - [x] Add `match` trace
  - [x] Add `add_deal` trace
- [x] Refactor match algorithm to make `moduleID` available one scope out 

This pull request adds a trace to deal making in the solver. It observes matching and reports deals that were made and errors where a deal could not be made.

Preview image:

CleanShot 2024-10-16 at 15 54 33@2x

### Task/Issue reference

Closes: #331 

### Test plan

Tracing over deal making includes a few spans:

- `solve`. The outermost parent span traces over the `solve` function.
- `get_matching_deals`. Child of the `solve` trace.
- `get_targeted_deal`. Child of `get_matching_deals`.
- `match`. Child of `get_matching_deals`.

Query for the `solve` span with:

```
{resource.service.name="solver" && name="solve"}
```

Add `span.deal_ids != "[]"` to filter out spans that did not match any deals:

```
{resource.service.name="solver" && name="solve" && span.deal_ids != "[]"}
```

Adding this filter will helps with testing because `solve` runs every ten seconds, but only one trace will make a match for a test job run.

The `get_matching_deals` span will appear as a child span. It can also be queried directly:

```
{resource.service.name="solver" && name="get_matching_deals" && span.job_offers != "[]"}
```

Filtering with `span.job_offers != "[]"` ignores spans without any matches because they lack a job offer to match on.

A `get_matching_deals` may contain one or more `match` spans. Each `match` span traces a match attempt between a job offer and a resource offer.

When using node targeting, a `get_targeted_deal` span will appear as a child of `get_matching_deals`. This can be queried directly:

```
{resource.service.name="solver" && name="get_targeted_deal"}
```

Use this query to observe targeting attempts that fail when the requested target is not available.

### Details

We use two different methods for tracing a function. When the function does not contain a child span, we wrap the call with `start` and `done` span events. For example:

https://github.com/Lilypad-Tech/lilypad/blob/fbc03828cf247002ad7f62e9c5a34a33cd920634/pkg/solver/matcher/matcher.go#L167-L175

When the function contains a child span, we simply call it passing the context and tracer. For example:

https://github.com/Lilypad-Tech/lilypad/blob/fbc03828cf247002ad7f62e9c5a34a33cd920634/pkg/solver/controller.go#L282

The two approaches give us visibility on each function call, either as a pair of events in a span or a child span.

Implements https://github.com/Lilypad-Tech/lilypad/issues/75

### DIFF
--- a/pkg/data/utils.go
+++ b/pkg/data/utils.go
@@ -33,12 +33,28 @@ func GetJobOfferID(offer JobOffer) (string, error) {
 	return CalculateCID(offer)
 }
 
+func GetJobOfferContainerIDs(jobOffers []JobOfferContainer) []string {
+	var ids []string
+	for _, offer := range jobOffers {
+		ids = append(ids, offer.ID)
+	}
+	return ids
+}
+
 func GetResourceOfferID(offer ResourceOffer) (string, error) {
 	offer.ID = ""
 	return CalculateCID(offer)
 }
 
 func GetResourceOfferIDs(resourceOffers []ResourceOffer) []string {
+	var ids []string
+	for _, offer := range resourceOffers {
+		ids = append(ids, offer.ID)
+	}
+	return ids
+}
+
+func GetResourceOfferContainerIDs(resourceOffers []ResourceOfferContainer) []string {
 	var ids []string
 	for _, offer := range resourceOffers {
 		ids = append(ids, offer.ID)
--- a/pkg/data/utils.go
+++ b/pkg/data/utils.go
@@ -38,6 +38,14 @@ func GetResourceOfferID(offer ResourceOffer) (string, error) {
 	return CalculateCID(offer)
 }
 
+func GetResourceOfferIDs(resourceOffers []ResourceOffer) []string {
+	var ids []string
+	for _, offer := range resourceOffers {
+		ids = append(ids, offer.ID)
+	}
+	return ids
+}
+
 func GetDealID(deal Deal) (string, error) {
 	deal.ID = ""
 	return CalculateCID(deal)
--- a/pkg/data/utils.go
+++ b/pkg/data/utils.go
@@ -43,6 +43,14 @@ func GetDealID(deal Deal) (string, error) {
 	return CalculateCID(deal)
 }
 
+func GetDealIDs(deals []Deal) []string {
+	var ids []string
+	for _, deal := range deals {
+		ids = append(ids, deal.ID)
+	}
+	return ids
+}
+
 func GetModuleID(module ModuleConfig) (string, error) {
 	return CalculateCID(module)
 }
--- a/pkg/solver/matcher/match.go
+++ b/pkg/solver/matcher/match.go
@@ -338,7 +338,7 @@ func logMatch(result matchResult) {
 	case marketPriceUnavailable:
 		log.Trace().
 			Str("resource offer", r.resourceOffer.ID).
-			Str("job offer", r.jobOffer.ID).
+			Str("pricing mode", string(r.resourceOffer.Mode)).
 			Msg(r.message())
 	case priceMismatch:
 		log.Trace().
--- a/pkg/solver/matcher/match.go
+++ b/pkg/solver/matcher/match.go
@@ -123,16 +123,17 @@ func matchOffers(
 		}
 	}
 
+	moduleID, err := data.GetModuleID(jobOffer.Module)
+	if err != nil {
+		return &moduleIDError{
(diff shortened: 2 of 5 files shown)


pkg/data/utils.go
```diff
@@ -33,16 +33,48 @@ func GetJobOfferID(offer JobOffer) (string, error) {
...
 	return CalculateCID(offer)
 }
 
+func GetResourceOfferIDs(resourceOffers []ResourceOffer) []string {
+	var ids []string
+	for _, offer := range resourceOffers {
+		ids = append(ids, offer.ID)
+	}
+	return ids
+}
```
bgins, line 55: It may be confusing that that `GetResourceOfferIDs` has a different behavior than `GetResourceOfferID`. Open to naming suggestions.

pkg/solver/matcher/match.go
```diff
@@ -51,35 +88,84 @@ type moduleIDError struct {
...
+	moduleID      string
 }
 
 func (_ priceMismatch) matched() bool { return false }
 func (_ priceMismatch) message() string {
 	return "fixed price job offer cannot afford resource offer"
 }
+func (result priceMismatch) attributes() []attribute.KeyValue {
+	// If the module instruction price is not specified, this lookup will use the zero-value of 0
+	moduleInstructionPrice := result.resourceOffer.ModulePricing[result.moduleID].InstructionPrice
```
bgins, line 152: We don't currently implement per module pricing, but this sets us up for it.

pkg/solver/matcher/matcher.go
```diff
@@ -18,31 +23,54 @@ func (a ListOfResourceOffers) Less(i, j int) bool {
...
 	})
 	if err != nil {
+		span.SetStatus(codes.Error, "get resource offers failed")
+		span.RecordError(err)
 		return nil, err
 	}
+	span.SetAttributes(attribute.KeyValue{
+		Key:   "resource_offers",
+		Value: attribute.StringSliceValue(data.GetResourceOfferContainerIDs(resourceOffers)),
+	})
```
bgins, line 49: In general, span attributes should be set when the span is created because they are generally considered static information. But we're setting them dynamically here as a part of the initialization leading up to the real work.

Setting these values as span attributes makes it easier to query spans with them.

pkg/solver/matcher/matcher.go
```diff
@@ -56,40 +84,72 @@ func GetMatchingDeals(
 		// loop over resource offers
 		matchingResourceOffers := []data.ResourceOffer{}
 		for _, resourceOffer := range resourceOffers {
+			_, matchSpan := tracer.Start(ctx, "match",
+				trace.WithAttributes(attribute.String("job_offer.id", jobOffer.ID),
+					attribute.String("resource_offer.id", resourceOffer.ID)),
+			)
```
bgins, line 90: Future work. Would like to refactor this part of the algorithm out into a function. The change would make reading this code easier and we could avoid defining an child span in the scope of its parent.

pkg/solver/matcher/match.go
```diff
@@ -123,16 +227,17 @@ func matchOffers(
...
 
+	moduleID, err := data.GetModuleID(jobOffer.Module)
+	if err != nil {
+		return &moduleIDError{
+			jobOffer:      jobOffer,
+			resourceOffer: resourceOffer,
+			err:           err,
+		}
+	}
+
```
walkah, line 238: why did you move this?

walkah (approved):
LGTM and local testing worked. One question, but otherwise great work! 🎉

pkg/solver/matcher/match.go
bgins: Moved it to access it in the price mismatch case:

https://github.com/Lilypad-Tech/lilypad/blob/60f45b41b0cdbf028e21025d24b3d06148a525be/pkg/solver/matcher/match.go#L269-L273

This price mismatch case is one scope out from where we were getting the module ID. Moving it avoids recomputing the module ID for this case.

PR #337, opened 2024-09-06
feat: POW submission batching

kelindi:
### Summary

- [ ] Adds the `pow_submit_url` variable to the toml files
- [ ] Modifies the submitWork function to no longer send the transaction on chain, and instead routes the tx to the new POW batching service

### Task/Issue reference

Closes: [Issue:287](https://github.com/Lilypad-Tech/internal/issues/287)


pkg/web3/bindings/pow/pow.go
```diff
@@ -827,6 +827,7 @@ func (_Pow *PowTransactorSession) RenounceOwnership() (*types.Transaction, error
 //
 // Solidity: function submitWork(uint256 nonce, string nodeId) returns()
 func (_Pow *PowTransactor) SubmitWork(opts *bind.TransactOpts, nonce *big.Int, nodeId string) (*types.Transaction, error) {
+	opts.NoSend = true
```
alvin-reyes, line 830: let's create a new submit work function here so we don't touch any of the old features. See my branch here https://github.com/Lilypad-Tech/lilypad/compare/feat/rpc-batch-ar?expand=1

pkg/web3/types.go
```diff
@@ -22,6 +22,7 @@ type Web3Options struct {
 	MediationAddress  string `json:"mediation_address" toml:"mediation_address"`
 	JobCreatorAddress string `json:"jobcreator_address" toml:"jobcreator_address"`
 	PowAddress        string `json:"pow_address" toml:"pow_address"`
+	PowSubmitURL      string `json:"pow_submit_url" toml:"pow_submit_url"`
```
alvin-reyes: let's rename to `PowBatchWsUrl`. Reason is, we want to make it generic if we end up batching other requests later on.

alvin-reyes:
Closing this since we don't need it. Let's re-open if otherwise.

PR #363, opened 2024-09-16
fix: exclude canceled job offers

walkah:
### Summary

This PR adds a `GetJobOffer` query filter to exclude cancelled jobs by default (can be overridden with `query.IncludeCancelled`. 

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/297

### Test plan

- Using main, run a targetted job against a non-existent address. Then run another job (you should see a bunch of "Job cancelled" errors).
- Do the same thing again with this PR - should be resolved.

### DIFF
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -183,6 +183,9 @@ func (solverServer *solverServer) getJobOffers(res corehttp.ResponseWriter, req 
 	if notMatched := req.URL.Query().Get("not_matched"); notMatched == "true" {
 		query.NotMatched = true
 	}
+	if includeCancelled := req.URL.Query().Get("include_cancelled"); includeCancelled == "true" {
+		query.IncludeCancelled = true
+	}
 	return solverServer.store.GetJobOffers(query)
 }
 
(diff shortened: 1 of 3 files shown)


narbs91 (approved):
Tested and LGTM!  Plus is spits out an error message too, small nit pick would be to add some context to the error message (i.e why was it cancelled) but functionally this change seems to work.  Did you also test with running the onchain job execution?

image

narbs91:
Also seeing this WRN popup now after a successful and unsuccessful job:

`2024-09-16T10:22:59-07:00 WRN pkg/system/context.go:66 > CleanupManager: Cleanup called again after already called`

walkah:
> 2024-09-16T10:22:59-07:00 WRN pkg/system/context.go:66 > CleanupManager: Cleanup called again after already called

yep, unrelated https://github.com/Lilypad-Tech/internal/issues/296

walkah:
> Did you also test with running the onchain job execution?

on chain jobs don't currently have a mechanism for targetting (i believe that would require a smart contract update)

PR #367, opened 2024-09-16
refactor: Update doOffersMatch to matchOffers

### LINKED ISSUE #364: Refactor Solver matcher `doOffersMatch`
Our current implementation of `doOffersMatch` returns a boolean. We would like to return richer information to support traces without adding a multiple span events throughout the function body.

We may also refactor the matcher implementation into a subpackage to better organize the Solver.

bgins:
### Summary

- [x] Refactor `doOffersMatch` to return a match result interface
- [x] Rename `doOffersMatch` to `matchOffers`
- [x] Refactor matcher implementation into a Solver subpackage
- [x] Add `matchOffers` unit tests

This pull request refactors the `doOffersMatch` check to return richer information about why a match was or was not made. The change is primarily in support our Solver deal making traces (#331), but also gives us more information on match results should we need it.

The addition of a `match.go` file was enough to crowd the `solver` package, so this PR also factors the matcher into a `matcher` subpackage.

The additional unit tests cover some cases that were missing from the initial set of unit tests.

### Task/Issue reference

Closes: #364 

### Test plan

Run the matcher unit tests from the project root.

```sh
go test ./pkg/solver/matcher/...
```

Run a job and run the integration tests to check that jobs still work.


pkg/solver/matcher/matcher.go
```diff
@@ -183,7 +79,10 @@ func getMatchingDeals(
 				continue
 			}
 
-			if doOffersMatch(resourceOffer.ResourceOffer, jobOffer.JobOffer) {
+			result := matchOffers(resourceOffer.ResourceOffer, jobOffer.JobOffer)
```
narbs91, line 82: So clean 🚀

pkg/solver/matcher/match.go
```diff
@@ -0,0 +1,256 @@
...
+	jobOffer      data.JobOffer
+}
+
+func (_ solverMismatch) matched() bool   { return false }
+func (_ solverMismatch) message() string { return "no matching solver" }
+
+// the most basic of matchers
+// basically just check if the resource offer >= job offer cpu, gpu & ram
+// if the job offer is zero then it will match any resource offer
+func matchOffers(
```
narbs91, line 103: Let's add some unit tests for this, i have a suspicion there might be a bug in this somewhere

narbs91 (approved):
LGTM!

PR #371, opened 2024-09-18
feat: fetch machine specs from bacalhau

walkah:
### Summary

This PR updates a resource provider's resource offers based on the current compute nodes available to bacalhau.

It also introduces some additional `MachineSpec` properties (even though jobs / matching don't take advantage yet):
- List of available GPUs per compute node (with vendor, name and vram).

**Note**: If the bacalhau orchestrator becomes unavailable, the resource provider will exit. (Would we rather just remove existing resource offers and give it a chance to come back online?)

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/289

### DIFF
--- a/pkg/executor/bacalhau/bacalhau.go
+++ b/pkg/executor/bacalhau/bacalhau.go
@@ -99,6 +99,31 @@ func (executor *BacalhauExecutor) IsAvailable() (bool, error) {
 	return true, nil
 }
 
+func (executor *BacalhauExecutor) GetMachineSpecs() ([]data.MachineSpec, error) {
+	var specs []data.MachineSpec
+	result, err := executor.bacalhauClient.getNodes()
+	if err != nil {
+		return specs, err
+	}
+
+	for _, node := range result.Nodes {
+		spec := data.MachineSpec{
+			CPU: int(node.Info.ComputeNodeInfo.MaxCapacity.CPU) * 1000, // convert float to "mili-CPU"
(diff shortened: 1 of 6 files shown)


pkg/executor/bacalhau/bacalhau.go
```diff
@@ -99,6 +99,32 @@ func (executor *BacalhauExecutor) IsAvailable() (bool, error) {
...
+func (executor *BacalhauExecutor) GetMachineSpecs() ([]data.MachineSpec, error) {
+	var specs []data.MachineSpec
+	result, err := executor.bacalhauClient.getNodes()
+	if err != nil {
+		return specs, err
+	}
+
+	for _, node := range result.Nodes {
+		spec := data.MachineSpec{
+			CPU:  int(node.Info.ComputeNodeInfo.MaxCapacity.CPU) * 1000, // convert float to "mili-CPU"
```
bgins:
```suggestion
			CPU:  int(node.Info.ComputeNodeInfo.MaxCapacity.CPU) * 1000, // convert float to "milli-CPU"
```
nit: Should this be `milli`?

bgins (approved):
Tested this out on macOS and it doesn't break anything. We should get a review from someone with a GPU to confirm the outputs from Bacalhau.

walkah:
This was tested by @noryev on his RP (4070ti) as well

PR #381, opened 2024-09-24
fix: resourceoffer updates

walkah:
### Summary

This PR tracks a timestamp in the resource provider controller for when we last reported our resource offers - since we don't need to update every 10 seconds. 

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/303


narbs91 (approved):
LGTM!

PR #384, opened 2024-09-24
chore: Add web3 sdk tracers

### LINKED ISSUE #376: Add web3 SDK tracer
We would like to implement tracing for web3 events to check on failing RPC endpoints. We instantiate the web3 SDK in the resource provider here:

https://github.com/Lilypad-Tech/lilypad/blob/4214994baf5d127ed78f00eb9a0162e162ef74c4/cmd/lilypad/resource-provider.go#L42-L45

Our other services also use `web3.NewContractSDK` to instantiate an SDK, so we will need to add a tracer to all services to pass to the function. We can use a noop tracer for services that do not trace yet.

bgins:
### Summary

- [x] Add tracers to web3 SDK instantiation
- [x] Replace tracerProvider with tracer on resource provider controller
- [x] Add `configureTelemetry` helper function
- [x] Add noop tracers to services that do not trace

This pull request prepares us to add resource provider web3 RPC connection traces (https://github.com/Lilypad-Tech/internal/issues/299). It adds a tracer to `NewContractSDK` which we will use for tracing in an upcoming PR.

### Task/Issue reference

Closes: #376 

### Test plan

Start the observability stack (https://github.com/Lilypad-Tech/observability) with:

```
./stack observe-up dev
```

Run a job and check that a `run_job` trace is recorded.

#### Noop tracers

Noop tracers were added to all services that call `NewContractSDK` but do not currently trace.

#### tracerProvider -> tracer

This PR takes a new approach to instantiating the `tracer`. We previously instantiated the tracer right before starting the `run_job` trace in the resource provider:

https://github.com/Lilypad-Tech/lilypad/blob/4ddf62dc3a340c771d4c416ad6e1359ddd9908c6/pkg/resourceprovider/controller.go#L419

The OpenTelemetry specification says a [tracer should uniquely identify](https://opentelemetry.io/docs/specs/otel/trace/api/#get-a-tracer) an instrumentation scope. An [instrumentation scope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) is:

> A logical unit of the application code with which the emitted telemetry can be associated. It is typically the developer’s choice to decide what denotes a reasonable instrumentation scope. The most common approach is to use the [instrumentation library](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-library) as the scope, however other scopes are also common, e.g. a module, a package, or a class can be chosen as the instrumentation scope.

After reading this in the spec, I think each of our services can be considered a "logical unit of the application code". 

This pull requests instantiates a single tracer for the resource provider and adds it to the resource provider controller and web3 SDK.

### DIFF
--- a/cmd/lilypad/resource-provider.go
+++ b/cmd/lilypad/resource-provider.go
@@ -70,7 +70,8 @@ func runResourceProvider(cmd *cobra.Command, options resourceprovider.ResourcePr
 	}
 	commandCtx.Cm.RegisterCallbackWithContext(telemetry.Shutdown)
 
-	resourceProviderService, err := resourceprovider.NewResourceProvider(options, web3SDK, executor, telemetry)
+	tracer := telemetry.TracerProvider.Tracer(system.GetOTelServiceName(system.ResourceProviderService))
+	resourceProviderService, err := resourceprovider.NewResourceProvider(options, web3SDK, executor, tracer)
 	if err != nil {
 		return err
 	}
--- a/cmd/lilypad/resource-provider.go
+++ b/cmd/lilypad/resource-provider.go
@@ -39,7 +39,30 @@ func runResourceProvider(cmd *cobra.Command, options resourceprovider.ResourcePr
 	commandCtx := system.NewCommandContext(cmd)
 	defer commandCtx.Cleanup()
 
-	web3SDK, err := web3.NewContractSDK(options.Web3)
+	// Extract address for telemetry
+	privateKey, err := web3.ParsePrivateKey(options.Web3.PrivateKey)
+	if err != nil {
+		return err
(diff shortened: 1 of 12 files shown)


narbs91 (approved):
LGTM

cmd/lilypad/resource-provider.go
```diff
@@ -55,22 +62,7 @@ func runResourceProvider(cmd *cobra.Command, options resourceprovider.ResourcePr
 		return err
 	}
 
-	tc := system.TelemetryConfig{
```
walkah, line 58: 🧹 🙌

walkah (approved):
LGTM! Runs cleanly locally

PR #390, opened 2024-09-26
feat: Add get_ethclient trace

bgins:
### Summary

- [x] Refactor `ethclient` dialing into a `getEthClient` function 
- [x] Add a `get_ethclient` trace over `getEthClient`
- [x] Handle error case when all RPC URL connections fail
- [x] Update log level for individual failures to warn

This pull request adds a trace where we connect to web3 RPC URLs when instantiating the web3 SDK.

Note that this trace will only be reported for the solver and resource providers because the other services currently have noop tracers.

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/299

### Test plan

Start the observability stack (https://github.com/Lilypad-Tech/observability) with:

```
./stack observe-up dev
```

Start the stack. Open the observability dashboard at `localhost:3000` and run the following TraceQL query:

```
{name="get_ethclient"}
```

Successful connections should be reported for the solver and resource provider.

Update the `WEB3_RPC_URL`:

https://github.com/Lilypad-Tech/lilypad/blob/b97815edf8a89f87e7edaf541d49bbb19206f1cd/.local.dev#L10

to 

```
WEB3_RPC_URL=ws://localhost:-1,ws://localhost:854
```

These changes will exercise two error conditions -- a malformed URL and a URL with an incorrect port. Both should fail and the `get_ethclient` span should be marked with an error because all URLs have failed.

Now update `WEB3_RPC_URL` to:

```
WEB3_RPC_URL=ws://localhost:854,ws://localhost:8548
```

An span event exception should be reported, but the span should be unmarked overall because the second URL should successfully connect.

### Details

Note that the traces only report the host to avoid sending sensitive data like tokens that might be included in the URL.

We may want to randomly select testnet RPC URLs to determine how reliably we can connect to them in the future.

### DIFF
--- a/pkg/web3/sdk.go
+++ b/pkg/web3/sdk.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
+	"fmt"
 	"math/big"
+	"net/url"
 	"strconv"
 	"strings"
 
@@ -22,6 +24,8 @@ import (
 	"github.com/lilypad-tech/lilypad/pkg/web3/bindings/token"
 	"github.com/lilypad-tech/lilypad/pkg/web3/bindings/users"
 	"github.com/rs/zerolog/log"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/trace"
 )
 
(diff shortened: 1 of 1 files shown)


walkah (approved):
LGTM and seems to work locally ✅

PR #396, opened 2024-10-02
fix: re-register resource offers after runJob

walkah:
### Summary

This is a follow up to https://github.com/Lilypad-Tech/lilypad/pull/381 that adds a call to re-register an RP's resource offer after the resource provider has successfully run a matched deal. The previous PR increased the control loop timeout between checks to 10 minutes (rather than 10 seconds) which creates unnecessary load on the solver. However (particularly in local testing/development) this creates a lag before your single RP becomes available again. 

*note* it also looks like it's been a while since we did a `go mod tidy` (hence the go.mod/sum noise)

### Test plan

Run a job locally, you should then see the solver register a new resource offer (` Solver solving deals=0 jobOffers=0 resourceOffers=1 `)  immediately after a job completes


bgins (approved):
Looks good! Tested locally and the resource offers are indeed updated after completing a job. 💯

PR #403, opened 2024-10-12
fix: lower default pricing and check LP balance

alvin-reyes:
### Summary

- [ ] Lower the instruction price to add resource offers
- [ ] Add a LP balance check 

### Task/Issue reference

Closes: [add_link_here](https://github.com/Lilypad-Tech/internal/issues/309)

### Test plan

Please describe how reviewers can test the changes in this pull request. If the test plan is challenging to explain in text alone, a short video demonstration may be included.

#### Positive tests
- Run the RP with an address ETH balance of > 0.0006 ETH 
- Run the RP with an address LP balance of > 0 LP
- Run the `GetBalance` and `GetLPBalance` unit tests

#### Negative tests
- Run the RP with an address ETH balance of 0 OR < 0.0006 ETH 
- Run the RP with an address LP balance of 0 or < 1 LP

## More details

- Key thing to note here is that we will need to add a new DealPricing property for LP and ETH as part of the job fee payment. I'll be putting adding that on a separate PR.

- The 0.0006 requirement here is based on the fees we pay per job. It is not an AVG, but just a educated guess. I'll have to create a new way of getting the fees (through indexing). Possibly a microservice to get the past 100 blocks, average the fee and use the value instead. I'll put this on a separate PR.


pkg/solver/controller.go
```diff
@@ -332,12 +332,23 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 		return nil, fmt.Errorf("failed to retrieve ETH balance for resource provider: %v", err)
 	}
 	// Convert InstructionPrice from ETH to Wei
-	requiredBalanceWei := web3.EtherToWei(float64(resourceOffer.DefaultPricing.InstructionPrice))
+	requiredBalanceWei := web3.EtherToWei(0.0006) // 0.0006 based on the required balance for a job
```
narbs91: [nit] Make the `0.0006` a variable so it's easier to track and change when we need to

pkg/web3/sdk.go
```diff
@@ -308,3 +312,30 @@ func (sdk *Web3SDK) GetBalance(address string) (*big.Int, error) {
...
+	erc20ABIObj, err := abi.JSON(strings.NewReader(erc20ABI))
+	if err != nil {
+		log.Printf("error parsing ABI: %s", err)
+	}
+	tokenContract := bind.NewBoundContract(lpToken, erc20ABIObj, client, client, client)
+
+	var out []interface{}
+	err = tokenContract.Call(nil, &out, "balanceOf", ethAddress)
+	if err != nil {
+		log.Printf("error calling balanceOf: %s", err)
```
narbs91:
```suggestion
		log.Error().Msgf("error calling balanceOf: %s", err)
```

pkg/web3/sdk.go
```diff
@@ -308,3 +312,30 @@ func (sdk *Web3SDK) GetBalance(address string) (*big.Int, error) {
...
+func (sdk *Web3SDK) GetLPBalance(address string) (*big.Int, error) {
+	// Convert the string address to common.Address
+	client := sdk.Client
+	ethAddress := common.HexToAddress(address)
+	lpToken := common.HexToAddress(sdk.Options.TokenAddress) // LP Token Address
+
+	// Get the balance using the converted address
+	erc20ABIObj, err := abi.JSON(strings.NewReader(erc20ABI))
+	if err != nil {
+		log.Printf("error parsing ABI: %s", err)
```
narbs91:
```suggestion
		log.Error().Msgf("error parsing ABI: %s", err)
```

pkg/web3/sdk_test.go
```diff
@@ -70,6 +73,18 @@ func TestGetBalance(t *testing.T) {
 	t.Logf("Balance: %d\n", balance)
 }
 
+func TestGetLPBalance(t *testing.T) {
```
narbs91, line 80: Maybe let's add another test for the case where there the address doesn't have enough LP

narbs91 (commented):
LGTM, just some minor comments

pkg/web3/sdk.go
alvin-reyes: resolved on the latest commit.

pkg/web3/sdk_test.go
alvin-reyes: resolved on the latest commit.

pkg/web3/sdk.go
alvin-reyes: resolved on the latest commit.

pkg/solver/controller.go
alvin-reyes: resolved on the latest commit.

pkg/web3/sdk.go
```diff
@@ -308,3 +312,30 @@ func (sdk *Web3SDK) GetBalance(address string) (*big.Int, error) {
...
+		log.Error().Msgf("error parsing ABI: %s", err)
+	}
+	tokenContract := bind.NewBoundContract(lpToken, erc20ABIObj, client, client, client)
+
+	var out []interface{}
+	err = tokenContract.Call(nil, &out, "balanceOf", ethAddress)
+	if err != nil {
+		log.Error().Msgf("error calling balanceOf: %s", err)
+	}
+	lpBalance := *abi.ConvertType(out[0], new(big.Int)).(*big.Int)
```
narbs91, line 337: I think the 'out[0]' will throw and out of bounds error here if we enter the err block above it.  Should we return if we're in the error block just above this line?

alvin-reyes: good catch. we should. resolved on the latest commit.

narbs91 (approved):

PR #408, opened 2024-10-18
feat: Add Lilypad version HTTP header

### LINKED ISSUE #406: Add version on HTTP requests
We would like to know what Lilypad version resource providers and job creators are running when they interact with solver and API servers. This information will help us to determine compatibility and gate access on deprecated versions where needed.

We could add an `x-` header with the current Lilypad version to all HTTP requests. This information would give us immediate visibility on all requests coming into the solver and API servers.

bgins:
### Summary

- [x] Add `X-Lilypad-Version` HTTP header
- [x] Log header on solver when receiving resource offers

The `X-Lilypad-Version` header is added to all HTTP requests across services. 

As a demonstration, this PR adds a solver log to report the version when a resource provider makes an add resource offer post. We can remove the log or keep it if it seems useful.

### Task/Issue reference

Closes: #406 

### Test plan

Build the binary with a placeholder version for testing:

```
go build -v -ldflags="-X 'github.com/lilypad-tech/lilypad/pkg/system.Version=v2.9.0' -w" .
```

Run the stack, using the binary for the resource provider and supplying our development-only private key:

```
WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a ./lilypad resource-provider --network dev
```

Check the solver logs for a "resource provider adding offer with version header v2.9.0" log.

### DIFF
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -5,6 +5,7 @@ import (
 	"crypto/ecdsa"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	stdlog "log"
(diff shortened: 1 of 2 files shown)


narbs91 (approved):
LGTM

PR #415, opened 2024-10-24
fix: add version info to RP Docker build

walkah:
### Summary

This changes the resource provider `Dockerfile` to take 2 additional build args: `VERSION` and `COMMIT_SHA` to provide to go build.  It also adds those variables to `./stack` build commands and the `release_docker` github workflow. 

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/327

### Test plan

You can test locally by running `./stack compose-build` (you should see the proper `go build` command run) you can then run the full stack locally w/ observability and you should see version info in traces.

### DIFF
--- a/stack
+++ b/stack
@@ -26,6 +26,7 @@ function compose-init() {
 function compose-build() {
   load-local-env
   compose-env
+  get-build-version
   docker compose -f ./docker/docker-compose.dev.yml build "$@"
 }
 
@@ -51,6 +52,14 @@ function load-local-env() {
       export $line
     fi
   done < .local.dev
+}
+
+############################################################################
+# Get build version info
+############################################################################
+function get-build-version() {
+  export VERSION=$(git rev-parse --abbrev-ref HEAD)
+  export COMMIT_SHA=$(git rev-parse HEAD)
 }
 
 ############################################################################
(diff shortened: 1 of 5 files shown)


stack
```diff
@@ -53,6 +54,14 @@ function load-local-env() {
   done < .local.dev
 }
 
+############################################################################
+# Get build version info
+############################################################################
+function get-build-version() {
+  export VERSION=$(git rev-parse --abbrev-ref HEAD)
```
bgins, line 61: It looks like this command returns the branch name:

```
»  git rev-parse --abbrev-ref HEAD
walkah/fix-add-version-to-docker-build
```

Perhaps we could use `git describe --tags --abbrev=0` to retrieve the latest tag? It won't be accurate for builds on `main` that aren't part of a release though.

bgins: Or possibly `git tag --points-at HEAD`. It returns a list of tags or an empty list if none, not sure if the output is ideal. 🤔

walkah: @bgins in the `release_docker` workflow, we're passing in the release tag name for VERSION. What's in `stack` would be intended for local dev, where I thought branch name made more sense. Does that help?

bgins: Yep, it does! Thanks for the explanation.

When we run `release_docker` on push to `main`, this will show the version as `main` right? I'd assume resource providers generally won't be running these images, but it's a consideration for parsing the version when we check it.

bgins (approved):
Looks good! ✨

stack
walkah: That's right, main will have a `main` version, with the corresponding git hash. Local docker builds will have branch name (plus git hash) and tagged releases will have tag / release name and git hash.

PR #416, opened 2024-10-24
fix: Handle accept result and download result errors

### LINKED ISSUE #414: Handle accept result and download result errors
Our job creator implementation silently drops accept result and download result errors:

https://github.com/Lilypad-Tech/lilypad/blob/b467d0fa36f9720038bf5714d7ce1409a375d2a8/pkg/jobcreator/controller.go#L316-L326

These errors are fatal, so they may have been ignored to avoid crashing the onchain job creator. However, we do want the crash and error reporting for CLI job runs.

Let's handle these errors short term, but for future work let's look into custom errors to differentiate a fatal and non-fatal errors. This section of the code also likely needs a way to address multiple errors that could occur in this loop.

bgins:
### Summary

- [x] Handle `acceptResult` errors
- [x] Handle `downloadResult` errors
- [x] Check for existing downloads before downloading results
- [x] Update error channel to non-blocking buffered channel 

We would like to handle these errors to avoid silently failing when running jobs from the CLI. Error reporting also supports observing this information in traces.

The unbuffered error channel blocks on the receiver which does not get set up before the send. The buffered channel will hold the message until the receiver is available to receive the message.

### Task/Issue reference

Closes: #414 

### Test plan

Run a few jobs. Check that everything still works.

Running more than one job exercises the download check. If we attempt to download again, we would observe a `file exists` error.

To test the error cases, return a temporary error from the `downloadResult` and `acceptResult` functions.

### Details (optional)

The new downloads check determines if we have already downloaded the results for a job. If the download path already exists, we skip the download. This check is necessary because our control loop checks for any completed deals and re-downloads the results. The solver may report completed deals for jobs whose results we already have.

### DIFF
--- a/pkg/jobcreator/controller.go
+++ b/pkg/jobcreator/controller.go
@@ -319,7 +319,11 @@ func (controller *JobCreatorController) checkResults() error {
 			// there is an error with the job
 			// accept anyway
 			// TODO: trigger mediation here
-			controller.acceptResult(dealContainer)
+			err := controller.acceptResult(dealContainer)
+			if err != nil {
+				controller.log.Error("failed to accept results", err)
+				return err
+			}
 		} else {
 			controller.downloadResult(dealContainer)
 		}
@@ -337,7 +341,11 @@ func (controller *JobCreatorController) downloadResult(dealContainer data.DealCo
 	controller.log.Debug("Downloaded results for job", solver.GetDownloadsFilePath(dealContainer.ID))
 
 	// TODO: activate the mediation check here
-	controller.acceptResult(dealContainer)
+	err = controller.acceptResult(dealContainer)
+	if err != nil {
+		controller.log.Error("failed to accept results", err)
+		return err
+	}
 
 	// work out if we should check or accept the results
 	// if controller.options.Mediation.CheckResultsPercentage >= rand.Intn(100) {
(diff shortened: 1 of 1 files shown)


pkg/jobcreator/controller.go
```diff
@@ -319,9 +322,23 @@ func (controller *JobCreatorController) checkResults() error {
...
+				controller.log.Error("failed to accept results", err)
+				return err
+			}
 		} else {
-			controller.downloadResult(dealContainer)
+			// We check for all completed deals, including deals whose results
+			// we have already downloaded. Check the download path and download
+			// if results do not exist.
+			downloadPath := solver.GetDownloadsFilePath(dealContainer.ID)
+			if _, err := os.Stat(downloadPath); errors.Is(err, fs.ErrNotExist) {
```
bgins, line 335: Common advice online recommends using `os.IsNotExist`, but the docs for this function recommend that new code should use `errors.Is(err, fs.ErrNotExist)`: https://pkg.go.dev/os#IsNotExist

pkg/jobcreator/controller.go
```diff
@@ -161,7 +164,7 @@ func (controller *JobCreatorController) subscribeToWeb3() error {
 }
 
 func (controller *JobCreatorController) Start(ctx context.Context, cm *system.CleanupManager) chan error {
-	errorChan := make(chan error)
+	errorChan := make(chan error, 1)
```
walkah, line 167: great catch on this! 🙏

walkah (approved):
🚢

PR #436, opened 2024-11-14
chore: ignore chain data directory

arsen3d:
### Summary
/chain-data/* directory does not need to be committed.


bgins (commented):
Haven't noticed a `chain-data` directory. How is this directory created?

bgins (approved):

PR #437, opened 2024-11-14
chore: disable warning message for gpu info

arsen3d:
### Summary
Adding this parameter removes:
```
WARNING: failed to determine nodes: open /sys/devices/system/node: no such file or directory
WARNING: failed to read int from file: open /sys/class/drm/card0/device/numa_node: no such file or directory
WARNING: failed to determine nodes: open /sys/devices/system/node: no such file or directory
WARNING: error parsing the pci address "vgem"
```
from output


bgins (commented):
Nice! Good catch to disable these warnings.

The change to `.gitignore` seems unrelated. Could we remove that from the PR?

bgins:
We might also want to rename the task here to `chore` instead of `fix`. Displaying a warning message is not a bug, the change here is more of a clean up chore.

bgins (approved):
Looks good! ✨

PR #438, opened 2024-11-14
fix: run bacalhau in a separate container

walkah:
### Summary

Previously, we bundled bacalhau and lilypad in a single container for resource providers. This meant that any crash/exit from lilypad would restart both bacalhau and lilypad. This could create lots of phone-home version checks for bacalhau. This PR also:

- Adds a `HEALTHCHECK` to the bacalhau Dockerfile so that we can properly wait for it to boot before starting the RP
- Moves local dev solver to port 8081 (to avoid conflicts with kubo HTTP gateway on 8080)

### Task/Issue reference

Closes: https://github.com/Lilypad-Tech/internal/issues/339

### Test plan

- Integration tests should still run (using `docker-compose.dev.yaml`). 
- For RPs, grab the latest `docker-compose.yaml` file and try to run. You should still be able to run jobs.


pkg/executor/bacalhau/bacalhau.go
```diff
@@ -208,7 +208,7 @@ func (executor *BacalhauExecutor) getJobID(
 	outputError := strings.Join(strings.Fields(strings.Join(splitOutputs[1:], " ")), " ")
 
 	if outputError != "" {
-		return "", fmt.Errorf("error running command %s -> %s, %s", deal.ID, outputError, runOutput)
+		log.Error().Msgf("error parsing output %s -> %s, %s", deal.ID, outputError, runOutput)
```
bgins, line 211:
```suggestion
		log.Error().Msgf("error found while parsing output %s -> %s, %s", deal.ID, outputError, runOutput)
```
Possible edit. It's not a parsing error, but an error we found while parsing.

walkah: Do you have thoughts about it being `ERR` vs `WARN`?

bgins: Tough call. On the one hand, the possible error we know about (around the version check) is more of a warning. On the other hand, could there be other errors that aren't a warning?

I'd say let's go with the common case we know about and use `WARN`. If we start seeing other errors, we can revisit, but this may not matter once we have upgraded to the latest Bacalhau.

bgins:
Wondering if we may want to update the solver `SERVER_PORT` to `8081` across the board for consistency. A few other places in the code that could be updated:

https://github.com/Lilypad-Tech/lilypad/blob/fe48d8cc3fb062708d3d573b73dcbfe14d7357b8/pkg/options/server.go#L14
https://github.com/Lilypad-Tech/lilypad/blob/fe48d8cc3fb062708d3d573b73dcbfe14d7357b8/docker/solver/Dockerfile#L25
https://github.com/Lilypad-Tech/lilypad/blob/fe48d8cc3fb062708d3d573b73dcbfe14d7357b8/hardhat/scripts/print-l2-config.ts#L40-L41
https://github.com/Lilypad-Tech/lilypad/blob/fe48d8cc3fb062708d3d573b73dcbfe14d7357b8/stack#L237

Outside of the code, we would need to update:
- Doppler `solver` projects
- Cloudflare tunnels

pkg/executor/bacalhau/bacalhau.go
walkah: yeah, once we get fully over to the bacalhau http api, a lot of this code become moot. 👍

walkah:
> Wondering if we may want to update the solver `SERVER_PORT` to `8081` across the board for consistency.

My thinking (and I can be convinced otherwise) was that just in local dev / CI there's potential for port conflicts with kubo on `8080` but in deployments, it's running in a more isolated context. I can definitely to the other updates though!

bgins:
> > Wondering if we may want to update the solver `SERVER_PORT` to `8081` across the board for consistency.
> 
> My thinking (and I can be convinced otherwise) was that just in local dev / CI there's potential for port conflicts with kubo on `8080` but in deployments, it's running in a more isolated context. I can definitely to the other updates though!

Yeah, that makes sense! Let's go with it.

bgins (approved):
Looks good! Integration tests and jobs running over here with the new setup. ✨

PR #450, opened 2024-11-25
feat: solver exponential backoff

### LINKED ISSUE #429: Add exponential backoff to websocket client
## General Description

Implement exponential backoff for the websocket client.

This change affects:
- Job Creation
- Resource Provider
- Mediation
- Solver

Our existing websocket client has a retry strategy where it attempts to connect every two seconds:

https://github.com/Lilypad-Tech/lilypad/blob/769deb7725d57fe5129f59d469982fba2ceb2f73/pkg/http/websocket_client.go#L17-L29

The repeated connection attempts can flood the solver with connection attempts at scale.

Let's implement exponential backoff when making the connection. We could start with a base interval of one second and an exponent of 1.5. Using an [exponential backoff calculator](http://exponentialbackoffcalculator.com/), this would look like:

In addition, we should set a maximum number of attempts before giving up and exiting the program. We should exit with a clear error message so the user can report an issue connecting to us.

---

The job creator, resource provider, and mediator connect to the solver over a websocket. This feature impacts the communication between each client and the solver.

PBillingsby:
### Summary

- [x] Add exponential backoff mechanism for connection retries
- [x] Improve detection of specific WebSocket closure errors
- [x] Add error propagation to ensure issues during WebSocket setup are logged 
- [x] Add mutexes (sync.Mutex) to synchronize access to the WebSocket connection during reconnection or ping operations
- [x] Change the signature of `ConnectWebSocket` to return an error alongside the response channel
- [x] Move ctx to start of params for `SolverClient`

### Task/Issue reference

Closes: #429 

### Test plan

1. Modify `pkg/http/websocket_server.go` to force an error in `HandleFunc`: 
```
r.HandleFunc(path, func(w http.ResponseWriter, r *http.Request) {
		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
		return
```
2. Start the stack
3. Observe the `job-creator` terminal for exponential backoff logs
```
ERR pkg/http/websocket_client.go:99 > WebSocket connection failed: websocket: bad handshake
Reconnecting in 2.200 seconds...
2024-11-25T11:11:28-05:00 DBG pkg/http/websocket_client.go:99 > WebSocket connection connecting: ws://localhost:8080/api/v1/ws?&Type=JobCreator&ID=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC
2024-11-25T11:11:28-05:00 ERR pkg/http/websocket_client.go:99 > WebSocket connection failed: websocket: bad handshake
Reconnecting in 3.640 seconds...
```
4. Cancel the solver service context to see the context cancellation handling:
```
2024-11-25T11:11:54-05:00 INF pkg/http/websocket_client.go:99 > Solver service appears to be down
2024-11-25T11:11:54-05:00 ERR pkg/solver/client.go:34 > Error in WebSocket connection. error="solver service unavailable: dial tcp [::1]:8080: connect: connection refused"
Error: solver service unavailable: dial tcp [::1]:8080: connect: connection refused
```

### DIFF
--- a/pkg/http/websocket_client.go
+++ b/pkg/http/websocket_client.go
@@ -5,8 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"math"
-	"net"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


pkg/http/websocket_client.go
```diff
@@ -2,76 +2,152 @@ package http
...
 	"time"
 
 	"github.com/gorilla/websocket"
 	"github.com/rs/zerolog/log"
 )
 
+const (
+	initialDelay = 1.0
+	maxAttempts  = 10
+	exponential  = 1.2
```
bgins, line 19: nit: Could we rename this to `exponent`?

pkg/http/websocket_client.go
```diff
@@ -2,76 +2,152 @@ package http
...
+			conn.Close()
+			close(responseCh)
+		}()
 		for {
-			messageType, p, err := conn.ReadMessage()
-			if err != nil {
-				errCh <- err
+			select {
+			case <-ctx.Done():
+				log.Info().Msg("Exiting readMessage loop due to context cancellation.")
```
bgins, line 80:
```suggestion
				log.Debug().Msg("Exiting readMessage loop due to context cancellation.")
```
This detail is a bit internal, would suggest dropping the log level to `debug`.

pkg/http/websocket_client.go
```diff
@@ -2,76 +2,152 @@ package http
...
+
 			log.Debug().Msgf("WebSocket connection connecting: %s", url)
 			conn, _, err := websocket.DefaultDialer.Dial(url, nil)
 			if err != nil {
-				log.Error().Msgf("WebSocket connection failed: %s\nReconnecting in 2 seconds...", err)
-				time.Sleep(2 * time.Second)
-				continue
+				switch {
+				case websocket.IsCloseError(err, websocket.CloseGoingAway, websocket.CloseAbnormalClosure):
+					log.Info().Msg("Solver service closed connection")
```
bgins, line 38: At the moment, we only use this helper function to connect to the solver. But we might re-use it in other contexts, so we may want to reword these errors to something like "Websocket server closed connection".

bgins: Also, we may want to log errors we where we will attempt again at the `warn` level.

pkg/http/websocket_client.go
```diff
@@ -2,76 +2,152 @@ package http
...
+					return nil, fmt.Errorf("solver connection closed: %w", err)
```
bgins, line 39: Will returning here cause us to give up on our connection attempts?

PR #458, opened 2024-11-28
fix: onchain job creator compatibility with full path to repo

arsen3d:
### Summary

The existed demo of on chain cowsay did not provide a way to pass in a repo path. Just repo name.
The update supports passing in a repo path, the same way we do it cli.


bgins:
@arsen3d how can we test this change? We have a client in `run-cowsay-onchain.ts` that has a `runCowsay` helper function, but it's not clear to me how this helper method works:

https://github.com/Lilypad-Tech/lilypad/blob/564fe64bd4c53c0fd172d2531e0ecdfbe18f548f/hardhat/scripts/run-cowsay-onchain.ts#L35-L37

Is there another way to test?

bgins (approved):
Looks good! 🙌

PR #464, opened 2024-12-05
fix: Fix resource offer JSON field name

bgins:
### Summary

- [x] Fix resource offer JSON field name

This change fixes a typo in the JSON serialization of `ResourceOfferContainer`. It changes the name from `job_offer` to `resource_offer`.

### Test plan

Start the stack. Run a job. Run the integration tests.

### Details

The `ResourceOfferContainer` is primarily used by the solver. Resource provider's request it when they are ensuring their offers: https://github.com/Lilypad-Tech/lilypad/blob/6469c5ec38a6afb9fbab6388272f59d518bc7ed5/pkg/resourceprovider/controller.go#L280

It should be safe to make this change without a migration plan, but would love a second look to make sure that's the case.


walkah (approved):
😂 🙌 🙏

PR #469, opened 2024-12-11
fix: Rename Result results_id JSON field name to deal_id

bgins:
### Summary

- [x] Change the JSON name for `Result.DealID` from `results_id` to `deal_id`
- [x] Add custom marshal and unmarshal `Result` functions for backwards compatibility
- [x] Add tests to verify marshal and unmarshal functions

The `Result` struct on `main` uses the JSON name `results_id` for its `DealID`. This pull request changes the JSON name to `deal_id`.

### Test plan

We have added property-based tests that check round trip serialization/deserialization and JSON backwards compatibility. Run the test suite and check that the new tests pass or run the tests directly:

```sh
go test pkg/data/data_test.go pkg/data/data.go -v --tags="unit" --count 1
```

Running a test job takes a few more steps than usual:

- Start the base services
- Start the solver on this pull request's branch (`bgins/fix-rename-results-id-field`)
- Start a resource provider from `main` or using a binary at `2.10.0`
- CLI run a job on `main` or using a binary at `2.10.0`

These steps check compatibility with a resource provider and job creator that use `results_id`.

### Details

Resource providers will continue to create results with the `results_id` field name until they have upgraded, and job creators will expect a `results_id` field until they have upgraded.

We have implemented a `Result` compatibility layer with custom serialization and deserialization:

- `MarshalJSON` adds the `results_id` field for clients that expect it
- `UnmarshalJSON` checks for a `results_id` field from older clients and assigns it to `DealID`

We can remove this compatibility layer once clients have upgraded to a version that uses `deal_id`.


walkah (approved):
Tests pass locally - and I was able to run a job from main.

Excellent test plan as usual. Thanks, @bgins !

Issue #424, opened 2024-11-01
Invalid character 'm' looking for beginning of value


bgins:
While investigating our error handling, I uncovered an issue reporting `HTTPError`s in response to posts to the solver.

Our `HTTPError` is defined here:

https://github.com/Lilypad-Tech/lilypad/blob/d30b6d81e7f315b37d8a2bf61210c6ef19a9877e/pkg/http/utils.go#L41-L44

It's a wrapper around an error message that let's us add a status code. 

When handling a POST, a handler may return an `HTTPError` or a plain `error`:

https://github.com/Lilypad-Tech/lilypad/blob/d30b6d81e7f315b37d8a2bf61210c6ef19a9877e/pkg/http/utils.go#L246-L257

When using an `HTTPError`, we take the provided status code or otherwise report a 500 `http.StatusInternalServerError` status code.

The set up for triggering this error is a bit contrived, but likely reveals an underlying issue where we will fail to correctly report errors.

Start the chain and solver. Comment out the line of code where we add a user header in our clients:

https://github.com/Lilypad-Tech/lilypad/blob/d30b6d81e7f315b37d8a2bf61210c6ef19a9877e/pkg/http/utils.go#L106

Start a resource provider. It should immediately fail with this error:

```
Error: invalid character 'm' looking for beginning of value
```

We handle this error when we check for the user header in the solver:

https://github.com/Lilypad-Tech/lilypad/blob/d30b6d81e7f315b37d8a2bf61210c6ef19a9877e/pkg/http/utils.go#L118-L124

Stop the solver. Change the status code in the user header check to `http.StatusInternalServerError`.

Run the resource provider again and it will not fail with the invalid character error. Instead, it will retry attempts to contact the solver and eventually fail when the max retries are reached.

You may also want to reduce the number of retries to speed up the process:

https://github.com/Lilypad-Tech/lilypad/blob/d30b6d81e7f315b37d8a2bf61210c6ef19a9877e/pkg/http/utils.go#L415

PR #472, opened 2024-12-16
feat: Add download files signature check

### LINKED ISSUE #473: Add signature check to dowload files route
## General Description

The download files route is currently exposed publicly. As a first step, let's check the signature of the job creator before allowing downloads.

This functionality affects the job creator and solver.

The job creator must sign requests to download files. The solver must check the signature before allowing the download to proceed.

bgins:
### Summary

- [x] Add signature check to download files route
- [x] Add headers to GET requests
- [x] Rename `GetAddressFromHeaders` to `CheckSignature`
- [x] Update error messages and comments 

This pull request adds a signature check to the download files route to ensure requests comes from the job creator that requested the job.

### Task/Issue reference

Closes: #473 

### Test plan

Start the stack. Run a job. Run the integration tests. Everything should work as expected.

We have also included a temporary commit to test an incorrect signature:

https://github.com/Lilypad-Tech/lilypad/blob/08a162524b158405b06e462ebc4631c300136cb0/pkg/http/utils.go#L323-L324

Comment line 323 and uncomment 324 to force a mismatch between the reported address and the one derived from the signature. Note that this test mechanism works because we do not check signatures on any other GET route (for now).

Note that error returned to the job creator when the signature fails does not correctly report an error message because we have yet to address: https://github.com/Lilypad-Tech/lilypad/issues/424

We will remove the temporary commit before merging this PR.

### Details

This pull request should be considered a minimal first step towards job output retrieval auth. In the future, we may want to implement decentralized auth to enable other actors to download job outputs.

### DIFF
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -319,6 +319,8 @@ func GetRequestBuffer(
 	if err != nil {
 		return nil, err
 	}
+	privateKey, err := web3.ParsePrivateKey(options.PrivateKey)
+	AddHeaders(req, privateKey, web3.GetAddress(privateKey).String())
 
 	resp, err := client.Do(req)
 	if err != nil {
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -109,12 +109,12 @@ func AddHeaders(
 	return nil
 }
 
-// this will use the client headers to ensure that a message was signed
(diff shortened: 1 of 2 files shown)


walkah (approved):
Thanks as always for the detailed test plan. Works for me! 🚢

PR #475, opened 2024-12-18
fix: Remove resource offers in DealNegotiating[0] state on disconnect

### LINKED ISSUE #467: Resource offer cleanup sometimes removes the wrong offers
## General Description

In our current implementation, the solver removes a resource offer when a resource provider disconnects:

https://github.com/Lilypad-Tech/lilypad/blob/da719da14890b2b3da2e4c725b6465759bde07b4/pkg/solver/server.go#L163-L173

The `removeResourceOfferByResourceProvider` implementation removes a single resource offer:

https://github.com/Lilypad-Tech/lilypad/blob/da719da14890b2b3da2e4c725b6465759bde07b4/pkg/solver/controller.go#L420-L423

We store the resource offers in a map:

https://github.com/Lilypad-Tech/lilypad/blob/da719da14890b2b3da2e4c725b6465759bde07b4/pkg/solver/store/memory/store.go#L15

When a job is complete, we mark the resource offer with a `ResultsAccepted[3]` state, but we do not remove it from the map.

Maps are unordered in Go, and a result `removeResourceOfferByResourceProvider` removes a random resource offer which may or may not be the active, unmatched resource offer.

In addition, resource providers may submit multiple resource offers. Cleaning up a single resource offer may not be enough in general.

The changes to fix this issue will affect the solver and resource providers.

## Describe the changes, and how this affects/ interacts with each system

We can fix this issue by removing all resource offers in a `DealNegotiating[0]` state. These are the resource offers that are unmatched, and we would like to remove them so they do not get matched when a resource provider is not connected to the solver.

We should avoid removing offers in any other state. Offers in a `DealAgreed[1]` and `ResultsSubmitted[2]` are in flight and we should be especially careful to preserve them.

bgins:
### Summary

- [x] Fix remove resource offers on resource provider disconnect
- [x] Rename `removeResourceOfferByResourceProvider` to `removeUnmatchedResourceOffers`

This pull request fixes a bug where we sometimes remove the wrong resource offer on disconnect. See https://github.com/Lilypad-Tech/lilypad/issues/467 for details.

The updated code removes resource offers in a unmatched `DealNegotiating[0]` state to avoid matching them when their resource provider is not connected. This approach has the additional benefit of removing multiple resource offers if the resource provider has offered them.

### Task/Issue reference

Closes: #467

### Test plan

We have included a temporary commit (`994bef2`) to demonstrate resource offer removal. The following steps are recommended for testing:

- Start the stack
- Stop the resource provider and observe the resource offers before and after. All offers should be removed because the one posted offer was in a `DealNegotiating[0]` state.
- Restart the resource provider and run a job
- Wait a moment, and stop the resource provider. The resource provider should have added a new resource offer in a `DealNegotiating[0]` state that gets removed, but the resource offer from running the job (in a `ResultsAccepted[3]`) state should be preserved.

Epic: https://github.com/Lilypad-Tech/internal/issues/367

### DIFF
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -404,7 +404,8 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 	return ret, nil
 }
 
-func (controller *SolverController) removeResourceOfferByResourceProvider(ID string) error {
+// Remove resource offers in an unmatched DealNegotiating[0] state
+func (controller *SolverController) removeUnmatchedResourceOffers(ID string) error {
 	controller.log.Info("remove resource offer", ID)
 	resourceOffers, err := controller.store.GetResourceOffers(store.GetResourceOffersQuery{
 		ResourceProvider: ID,
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -405,10 +405,10 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 }
 
 // Remove resource offers in an unmatched DealNegotiating[0] state
-func (controller *SolverController) removeUnmatchedResourceOffers(ID string) error {
-	controller.log.Info("remove resource offer", ID)
+func (controller *SolverController) removeUnmatchedResourceOffers(resourceProviderID string) error {
+	controller.log.Info("remove resource offer", resourceProviderID)
 	resourceOffers, err := controller.store.GetResourceOffers(store.GetResourceOffersQuery{
-		ResourceProvider: ID,
+		ResourceProvider: resourceProviderID,
 	})
 	if err != nil {
 		return err
@@ -419,7 +419,7 @@ func (controller *SolverController) removeUnmatchedResourceOffers(ID string) err
 			err = controller.store.RemoveResourceOffer(offer.ID)
 			if err != nil {
 				controller.log.Error("remove resource offer failed",
-					fmt.Errorf("resource provider: %s, offer ID: %s, error: %s", ID, offer.ID, err))
+					fmt.Errorf("resource provider: %s, offer ID: %s, error: %s", resourceProviderID, offer.ID, err))
 			}
 		}
 	}
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -418,7 +418,8 @@ func (controller *SolverController) removeUnmatchedResourceOffers(ID string) err
 		if offer.State == 0 {
 			err = controller.store.RemoveResourceOffer(offer.ID)
 			if err != nil {
-				controller.log.Error("remove resource offer failed: %s", err)
+				controller.log.Error("remove resource offer failed",
+					fmt.Errorf("resource provider: %s, offer ID: %s, error: %s", ID, offer.ID, err))
 			}
 		}
 	}
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -413,13 +413,13 @@ func (controller *SolverController) removeResourceOfferByResourceProvider(ID str
 		return err
 	}
 
-	if len(resourceOffers) == 0 {
-		return nil
-	}
-
-	err = controller.store.RemoveResourceOffer(resourceOffers[0].ID)
-	if err != nil {
-		return err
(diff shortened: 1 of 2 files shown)


pkg/solver/controller.go
```diff
@@ -404,23 +404,35 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
...
+func (controller *SolverController) removeUnmatchedResourceOffers(ID string) error {
 	controller.log.Info("remove resource offer", ID)
 	resourceOffers, err := controller.store.GetResourceOffers(store.GetResourceOffersQuery{
 		ResourceProvider: ID,
 	})
 	if err != nil {
 		return err
 	}
+	// TODO Remove before merge
+	fmt.Printf("+++ resource offers before removal: %+v\n", resourceOffers)
```
narbs91: [nit] Reminder to remove once we're ready to merge

pkg/solver/controller.go
```diff
@@ -404,23 +404,35 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
...
 
-	if len(resourceOffers) == 0 {
-		return nil
+	for _, offer := range resourceOffers {
+		if offer.State == 0 {
+			err = controller.store.RemoveResourceOffer(offer.ID)
+			if err != nil {
+				controller.log.Error("remove resource offer failed: %s", err)
```
narbs91: Perhaps we should log some context around the resource offer to help debug issues

pkg/solver/controller.go
```diff
@@ -404,23 +404,35 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 	return ret, nil
 }
 
-func (controller *SolverController) removeResourceOfferByResourceProvider(ID string) error {
+// Remove resource offers in an unmatched DealNegotiating[0] state
+func (controller *SolverController) removeUnmatchedResourceOffers(ID string) error {
```
narbs91: [nit] lets change `ID` to `resourceProviderID` for more clarity cause I got confused at first and though this was supposed to be the resource offer id 😅

pkg/solver/controller.go
```diff
@@ -404,23 +404,35 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
...
+				controller.log.Error("remove resource offer failed: %s", err)
+			}
+		}
 	}
 
-	err = controller.store.RemoveResourceOffer(resourceOffers[0].ID)
+	// TODO Remove before merge
+	offersAfter, err := controller.store.GetResourceOffers(store.GetResourceOffersQuery{
+		ResourceProvider: ID,
+	})
```
narbs91: [nit] Reminder to remove once we're ready to merge

pkg/solver/controller.go
```diff
@@ -404,23 +404,35 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
...
 	if err != nil {
 		return err
 	}
+	fmt.Printf("+++ resource offers after removal: %+v\n", offersAfter)
```
narbs91: [nit] Reminder to remove once we're ready to merge

pkg/solver/controller.go
bgins: Yeah great idea! Added the resource provider ID and the ID of the offer that could not be removed.

pkg/solver/controller.go
bgins: Yeah totally. We lost some semantic meaning renaming the function. Pushed an update change `ID` to `resourceProviderID`.

pkg/solver/controller.go
narbs91: Are you still planning to leave these print statements around @bgins ?

bgins: Ah nope, planning to remove those. Are you done with your review except that? Can remove if so.

narbs91: Yep, everything else looks great!

narbs91 (approved):

pkg/solver/controller.go
bgins: Alright great! Removed them, waiting for the tests and will merge. ✨

PR #476, opened 2024-12-18
chore: Cleanup run command output

bgins:
### Summary

- [x] Remove `ipfs.io` URL from run output
- [x] Add data ID to run output

We will stop storing results in IPFS soon, and the IPFS URL will no longer be valid. It's still useful to view the job outputs CID, so this PR reports that as a Data ID.

### Test plan

Start the stack. Run a job using the CLI. Check the output at the end.

### Details

CleanShot 2024-12-18 at 11 24 41@2x


logan-wrld (approved):
Screenshot from 2024-12-18 18-07-54
Tested and everything looks great! @bgins

narbs91 (approved):

PR #480, opened 2025-01-08
chore: Improve get solver URL error message

bgins:
### Summary

- [x] Improve `GetSolverUrl` error message when solver URL not found onchain

When attempting to discover the solver URL, we have been reporting a "GetUser" error, which aligns with the code but is confusing for users. This pull request improves the error message.

### DIFF
--- a/pkg/web3/api.go
+++ b/pkg/web3/api.go
@@ -90,8 +90,7 @@ func (sdk *Web3SDK) GetSolverUrl(address string) (string, error) {
 		common.HexToAddress(address),
 	)
 	if err != nil {
-		log.Error().Msgf("GetUser error")
-		log.Error().Msgf("error: %s", err)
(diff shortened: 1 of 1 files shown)


pkg/web3/api.go
```diff
@@ -90,8 +90,7 @@ func (sdk *Web3SDK) GetSolverUrl(address string) (string, error) {
 		common.HexToAddress(address),
 	)
 	if err != nil {
-		log.Error().Msgf("GetUser error")
-		log.Error().Msgf("error: %s", err)
+		log.Error().Msgf("Failed to discover to solver URL: %s", err)
```
narbs91: [nit] small thing but is there an extra `to` between discover and solver?

bgins: Ah yeah, good catch 💯. Fixed it.

narbs91 (approved):
LGTM!

PR #489, opened 2025-01-16
fix: docker setup for RPs

walkah:
### Summary

A few small fixes for the docker configuration for Resource Providers:

- moves most of the build arg / env values to the proper build stage
- adds a service_healthy check to make sure bacalhau has started before starting the RP

### DIFF
--- a/docker/resource-provider/Dockerfile
+++ b/docker/resource-provider/Dockerfile
@@ -2,15 +2,8 @@ ARG COMPUTE_MODE=gpu
 
 FROM golang:latest AS base
 WORKDIR /usr/src/app
-ARG NETWORK=testnet
 ARG VERSION
 ARG COMMIT_SHA
-
-# Default environment variables
-ENV LOG_LEVEL=info
-ENV BACALHAU_API_HOST="localhost"
-ENV WEB3_PRIVATE_KEY=""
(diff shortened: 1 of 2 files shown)


docker/resource-provider/Dockerfile
```diff
@@ -28,8 +21,14 @@ RUN go build -v -ldflags="-X 'github.com/lilypad-tech/lilypad/pkg/system.Version
 ENV DISABLE_POW=true
 
 FROM build-$COMPUTE_MODE AS final
-RUN mv lilypad /usr/local/bin
+# Default environment variables
+ARG NETWORK=testnet
+ARG DISABLE_TELEMETRY=false
```
narbs91, line 26: This one was previously an ENV, but now an ARG, is that on purpose?

walkah: > This one was previously an ENV, but now an ARG, is that on purpose?

@narbs91 good catch! and yes, it's intentional - we "bake" that in at build-time to the `run` script (see a few lines below) so there's not actually any value having it as `ENV` (something to clean up later).

narbs91: Ah gotcha, good stuff!

narbs91 (approved):

Issue #486, opened 2025-01-14
Add validation token endpoint


bgins:
## General Description

Add a solver endpoint that issues JWT authorization tokens for our upcoming validation service.

## Which system(s) or functionality does this affect

This functionality affects the solver and validation.

## Describe the changes, and how this affects/ interacts with each system.

The solver will issue a JWT that authorizes clients to log data with the validation service.

PR #495, opened 2025-01-22
feat: Add validation token key ID config

bgins:
### Summary

- [x] Add `SERVER_VALIDATION_TOKEN_KID` CLI option and environment variable

We would like to make the Key ID JWT header configurable to set a different key ID per each network.

### Task/Issue reference

Amends: https://github.com/Lilypad-Tech/lilypad/issues/486

### Test plan

We have included a temporary commit (1f3f971e95b496bb1cca43760291b196026a59d7) to show a client request for a token. This temporary commit will be removed before merging this pull request.

### Details

We have added a `SERVER_VALIDATION_TOKEN_KID` solver server option. `SERVER_VALIDATION_TOKEN_KID` is required and does not have a default value.

The options can be configured through environment variables or CLI options:

```sh
./stack solver --server-validation-token-kid somekid
SERVER_VALIDATION_TOKEN_KID=somekid ./stack solver
```

Note that the `SERVER_VALIDATION_TOKEN_KID` is overridden by the key ID we have set for local development:

https://github.com/Lilypad-Tech/lilypad/blob/1f3f971e95b496bb1cca43760291b196026a59d7/stack#L211

Comment this line to test.

### Related issues or PRs

Epic: https://www.notion.so/lilypadnetwork/MVP-Validation-176155da99b5801ebeffc417f6b270c7

### DIFF
--- a/stack
+++ b/stack
@@ -208,6 +208,7 @@ function solver() {
   export STORE_CONN_STR=postgres://postgres:postgres@localhost:5432/solver-db?sslmode=disable
   export STORE_GORM_LOG_LEVEL=silent
   export SERVER_VALIDATION_TOKEN_SECRET=912dd001a6613632c066ca10a19254430db2986a84612882a18f838a6360880e
+  export SERVER_VALIDATION_TOKEN_KID=key-dev
   go run . solver --network dev "$@"
 }
 
--- a/pkg/http/types.go
+++ b/pkg/http/types.go
@@ -11,6 +11,7 @@ type ServerOptions struct {
 type AccessControlOptions struct {
 	ValidationTokenSecret     string
 	ValidationTokenExpiration int
+	ValidationTokenKid        string
 }
 
 type ValidationToken struct {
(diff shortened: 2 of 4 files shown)


stack
```diff
@@ -208,6 +208,7 @@ function solver() {
   export STORE_CONN_STR=postgres://postgres:postgres@localhost:5432/solver-db?sslmode=disable
   export STORE_GORM_LOG_LEVEL=silent
   export SERVER_VALIDATION_TOKEN_SECRET=912dd001a6613632c066ca10a19254430db2986a84612882a18f838a6360880e
+  export SERVER_VALIDATION_TOKEN_KID=key-dev
```
bgins, line 211: Thought that tagging the key IDs by network might be a good idea, but open to another naming convention.

hollygrimm: I agree it's much better, I've updated the jwt-auth to use `key-local`, `key-dev` (using doppler secret in devnet), and `key-test` (using doppler secret in testnet). Thanks Brian!

hollygrimm (approved):

stack
bgins: Yeah, having the names will be nice for debugging the tokens! 

We may want to keep `key-dev` for local development to match our naming convention in other places. The names we use for networks are `dev`, `devnet`, `demonet`, and `testnet`. Networks are selected at the CLI using these names, for example `--network devnet` or `--network testnet`, which load an embedded config with a matching name from: https://github.com/Lilypad-Tech/lilypad/tree/main/pkg/options/configs

PR #496, opened 2025-01-23
chore: Disable external dependencies for local dev

bgins:
### Summary

- [x] Disable API calls in local development
- [x] Disable telemetry in local development

The API and telemetry should not be required by default for local development, but they can be enabled when integrating with them.

### Test plan

Check that the instructions in https://github.com/Lilypad-Tech/lilypad/blob/main/LOCAL_DEVELOPMENT.md work without running the API or observability servers.

### Details

The API and telemetry can be enabled in local development.

Enable API calls by setting the `--api-host` CLI option. For example:

```
./stack solver --api-host http://localhost:8002/
```

Enable telemetry by setting the `--disable-telemetry` CLI option to `false`. For example:

```sh
./stack resource-provider --disable-telemetry=false
```

### DIFF
--- a/.local.dev
+++ b/.local.dev
@@ -1,12 +1,13 @@
-# Running the Lilypad application locally depends on the `.local.dev` file for secrets injection. 
-# This file contains a series of private keys (with no funds on them) that are ONLY meant to be used 
-# for testing this app locally. You are free to replace these keys with your own if you wish; however, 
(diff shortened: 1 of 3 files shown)


walkah (approved):
works as expected locally! 👍

PBillingsby:
Works like a charm @bgins !

PR #497, opened 2025-01-23
feat: Add matching sorts

### LINKED ISSUE #492: Adding sorting to solver matching algorithm
## General Description

The solver matches job offers and resource offers by a machine spec and statistics constraints. In addition, the matching algorithm may sort job offers and resource offers before, during, or after constraints matching.

We will add the following sorts to the initial version of the matching algorithm:

- Job offers are sorted oldest first
- When considering a job offer in fixed pricing mode, sort resource offers by oldest offer first
- When considering a job offer in market pricing mode, sort resource offers by cheapest offer first
  - If multiple resource offers offer the same cheap price, sort the resource offers again oldest first

The solver is affected directly. The matching algorithm has indirect consequences on resource providers and job creators.

The implementation will be entirely in the solver. We will add sorts once defined where appropriate in the matching code.

bgins:
### Summary

- [x] Update `GetJobOffers` store method with oldest first sort query
- [x] Update `GetResourceOffers` store method with oldest first query
- [x] Add pricing and age sorting preferences to matching algorithm
  - [x] Update matcher to retrieve offers oldest first from store
  - [x] Add `isCheaperOrOlder` match helper
    - [x] Add `TestIsCheaperOrOlder` unit test
  - [x] Add matching and sorting integration test
- [x] Add `matchOffers` benchmark
  - [x] Add `bench` GitHub action
  - [x] Document benchmarks in README and local development guide

In addition, it makes supporting and refactoring changes:

- [x] Pass solver service logger to matcher
- [x] Move `matcher_test.go` test cases to `pkg/solver/matcher/testing.go` for reuse by the benchmark
- [x] Move `store_test.go` store helpers to `pkg/solver/testing.go` for reuse by the matching and sorting integration test
- [x] Return match result values instead of refs from `matchResult` (fix for `logMatch` logger)

This PR updates the matching algorithm with pricing and age sorting preferences. In addition, we add new test coverage, including relative benchmarks to report performance and alert on performance degradation.

### Task/Issue reference

Closes: #492 

### Test plan

**Unit tests**

We have added a `TestIsCheaperOrOlder` unit test to test our `isCheaperOrOlder` match helper. The test runs with the other unit tests or alone with:

```
go test -tags="unit" -v -count 1 -run TestIsCheaperOrOlder ./.../matcher
```
**Integration tests**

We have extended our store integration tests to cover our get oldest offer queries. Run the store integration tests:

```
go test -tags="integration,solver" ./.../store -v -count 1
```

The new tests are `TestJobOfferSort` and `TestResourceOfferSort`.

We have added a new matching and sorting integration test to the solver. Run the test with:

```
go test -tags="integration,solver" ./.../solver -v -count 1
```

Run all of the integration tests with:

```
./stack integration-tests-solver
```

**Benchmarks**

Our first benchmark has arrived! We have added documentation describing the benchmark in the local development guide and a note celebrating them in our README. 🎉 

The benchmark measures the performance of `matchOffers`. Run it with:

```
./stack benchmarks-solver
```

Our primary goal is relative benchmarking in CI. Aside from running in a isolated GitHub action, we make no attempts to control the environment where the benchmark runs. We want to know if a change has degraded performance, but we are not too concerned with accuracy.

The new `bench` GitHub action runs the benchmark. We get some nice reporting and alerting from `benchmark-action/github-action-benchmark`:

- Commit comment alerts when a commit degrades performance. For example: https://github.com/Lilypad-Tech/lilypad/commit/249cca9fe568ecfc6f04813b48ad46ccc0c76258#commitcomment-151701246.
- GitHub action summaries for the `bench` action. For example: https://github.com/Lilypad-Tech/lilypad/actions/runs/13023788174/attempts/1#summary-36329383772

In addition, when we push to `main`, the action publishes the benchmark results to a historical chart: https://lilypad-tech.github.io/lilypad/dev/bench/ (some sample data from commits on this PR, can clean up later)

### Details

The pricing and age sorting preferences use the following criteria:

- When considering a job offer in fixed pricing mode, sort resource offers by oldest offer first
- When considering a job offer in market pricing mode, sort resource offers by cheapest offer first
  - If multiple resource offers offer the same cheap price, we sort the resource offers again oldest first

In practice, we request job offers and resource offers from the database sorted oldest first. For each job offer, we build a list of matching resource offers. We apply the pricing preference over the matched resource offers:

- If the job offer uses fixed pricing, select the first (oldest) resource offer that meets the price requirement
- If the job offer uses market pricing, select the cheapest offer and break ties with age

We may be able to optimize in the future by retrieving resource offers cheapest first from the database, but we believe this to be a good starting point with room to grow. 😄 

**Establishing offer age**

When we query the database store, we sort oldest first relying on the database engine to sort for us. We don't have this option for the memory store, so we use the job offer and resource offer `CreatedAt` field.

Note that this implementation likely makes the memory store unsuitable for production use cases. A job creator or resource provider could forge timestamps to game the system.

### Related issues or PRs

Epic: https://github.com/Lilypad-Tech/internal/issues/367


pkg/solver/matcher/match.go
```diff
@@ -330,15 +330,15 @@ func matchOffers(
 
 	// we don't currently support market priced resource offers
 	if resourceOffer.Mode == data.MarketPrice {
```
narbs91, line 332: [q] Would there be a case with the `jobOffer.Mode` be MarketPrice?

narbs91, line 333: [q] Would there be a case with the jobOffer.Mode be MarketPrice?

pkg/solver/matcher/match.go
```diff
@@ -348,20 +348,20 @@ func matchOffers(
 
 	mutualMediators := data.GetMutualServices(resourceOffer.Services.Mediator, jobOffer.Services.Mediator)
 	if len(mutualMediators) == 0 {
-		return &mediatorMismatch{
+		return mediatorMismatch{
```
narbs91, line 353: [nit] Let's add a TODO here to rename mediator to validator

pkg/solver/matcher/matcher.go
```diff
@@ -36,10 +27,11 @@ func GetMatchingDeals(
...
-	// Get resource offers
+	// Get resource offers oldest first
 	span.AddEvent("db.get_resource_offers.start")
 	resourceOffers, err := db.GetResourceOffers(store.GetResourceOffersQuery{
-		NotMatched: true,
+		NotMatched:       true,
+		OrderOldestFirst: true,
 	})
 	if err != nil {
 		span.SetStatus(codes.Error, "get resource offers failed")
```
narbs91, line 41: Let's add a log message here around the failure

pkg/solver/matcher/matcher.go
```diff
@@ -52,10 +44,11 @@ func GetMatchingDeals(
...
-	// Get job offers
+	// Get job offers oldest first
 	span.AddEvent("db.get_job_offers.start")
 	jobOffers, err := db.GetJobOffers(store.GetJobOffersQuery{
-		NotMatched: true,
+		NotMatched:       true,
+		OrderOldestFirst: true,
 	})
 	if err != nil {
 		span.SetStatus(codes.Error, "get job offers failed")
```
narbs91, line 59: Ditto about adding a log message here around the error

pkg/solver/matcher/matcher.go
```diff
@@ -101,27 +93,30 @@ func GetMatchingDeals(
 			}
 			matchSpan.AddEvent("db.get_match_decision.done")
```
narbs91, line 100: A potential optimization for later but on line 88 we are calling `db.GetMatchDecision(resourceOffer.ID, jobOffer.ID)` in each step of the loop.  Perhaps we can change this to instead get called as `GetMatchDecisions()` and have it hoisted just before the loop starts.  It's more to save on multiple network calls to the DB and have the code sift through the decisions but I think it should speed it up a bit assuming we have the indexes on the DB properly lined up

pkg/solver/matcher/matcher.go
```diff
@@ -136,49 +131,74 @@ func GetMatchingDeals(
...
+				for _, matchingOffer := range matchingResourceOffers {
+					addDealID := ""
+					if selectedResourceOffer.ID == matchingOffer.ID {
+						addDealID = deal.ID
+					}
+
+					// All match decisions had matching resource offers, set the result to true.
+					// The match decision has a deal ID if it's resource offer was selected.
+					span.AddEvent("add_match_decision.start")
+					_, err := db.AddMatchDecision(matchingOffer.ID, jobOffer.ID, addDealID, true)
```
narbs91: I would suggest that instead of firing off multiple create requests we instead take advantage of gorms batch upload capability.  So how it would work is that you'd create an array of decisions and push them one by one in this loop, then at the end of the loop you call a function to do a bulk upload.

You'd need to add another function in the db store that looks something like this:

```go
func (store *SolverStoreDatabase) AddBulkMatchDecisions(decisions []*model.MatchDecisions) error {
	// Split the records into batches to be inserted at once starting with a baseline of 100
	batchSize := 100

	return p.DB.Transaction(func(tx *gorm.DB) error {
		if err := tx.CreateInBatches(decisions, batchSize).Error; err != nil {
			log.Printf("...something bad happened...", err)

			return err
		}
		return nil
	})
}```

File: pkg/solver/matcher/matcher.go [hunk 7]
```diff
@@ -136,49 +131,74 @@ func GetMatchingDeals(
...
+					if err != nil {
+						span.SetStatus(codes.Error, "unable to add match decision")
+						span.RecordError(err)
+						return nil, err
+					}
+					span.AddEvent("add_match_decision.done")
```
narbs91: Maybe have this as it's own private function to shrink the size of the main function and help readability?

pkg/solver/matcher/matcher.go
```diff
@@ -84,7 +76,7 @@ func GetMatchingDeals(
 			continue
 		}
 
-		// loop over resource offers
+		// Build list of matching resource offers
 		matchingResourceOffers := []data.ResourceOffer{}
 		for _, resourceOffer := range resourceOffers {
```
narbs91, line 86: Maybe we can have this loop be a private function that does all the magic and returns the populated `matchingResourceOffers` array to help shrink down the overall function size and help readability

pkg/solver/matcher/matcher.go
narbs91: Also inside the loop we want to have a log when a match is made (though this might flood the log file) or have a log at the end of the loop to signify how many matches were made

pkg/solver/matcher/matcher.go
```diff
@@ -136,49 +131,74 @@ func GetMatchingDeals(
...
+					if selectedResourceOffer.ID == matchingOffer.ID {
+						addDealID = deal.ID
+					}
+
+					// All match decisions had matching resource offers, set the result to true.
+					// The match decision has a deal ID if it's resource offer was selected.
+					span.AddEvent("add_match_decision.start")
+					_, err := db.AddMatchDecision(matchingOffer.ID, jobOffer.ID, addDealID, true)
+					if err != nil {
+						span.SetStatus(codes.Error, "unable to add match decision")
```
narbs91: Should have a log an error log message here

narbs91 (commented):
Great stuff @bgins! Left some comments, let me know if you want to chat through any of them

pkg/solver/matcher/match.go
bgins: Yes, job offers are market priced by default: https://github.com/Lilypad-Tech/lilypad/blob/a6ddf89d6e5bc2814c393e1c5667a3208ec10eee/pkg/options/job-creator.go#L33

I expect this will be the common case at first. We will not immediately expose fixed pricing to job creators through the API.

pkg/solver/matcher/match.go
bgins: Done ✅ . Added it to the `mediatorMismatch` struct definition as well: 6f816ab55f6bdb038a11421ee121225e2e3d50f8

pkg/solver/matcher/matcher.go
bgins: Done: c56ccfb70bc00117d0239c5130ee80d024a99af1

Note that we weren't passing the solver service logger to `GetMatchingDeals`, so went ahead and added that here: ae7755acff280bc52563b5eb982a1e93986c0b85. Also added a note about this change to the PR description.

pkg/solver/matcher/matcher.go
bgins: Done: c56ccfb70bc00117d0239c5130ee80d024a99af1

pkg/solver/matcher/matcher.go
bgins: Done: c56ccfb70bc00117d0239c5130ee80d024a99af1

pkg/solver/matcher/matcher.go
bgins: Opened an issue and referenced this optimization for future work: https://github.com/Lilypad-Tech/lilypad/issues/503

pkg/solver/matcher/matcher.go
bgins: Opened an issue and referenced this optimization for future work: https://github.com/Lilypad-Tech/lilypad/issues/503

pkg/solver/matcher/matcher.go
bgins: Refactored this loop into a `addMatchDecisions` helper function: 5e99321b70604443fd213033d765eb7c037f1764

bgins: And added the logging here: 2a57b7b528394f4b8dbeb14e3238e1744c01e2ed

pkg/solver/matcher/matcher.go
bgins: Let's hold off on this one for now. Refactoring into private functions is a bit noisy at the moment because we need to pass the `db`, `tracer`, `log`, etc to each function that uses them.

We can improve this situation by adding a `matcher` struct and implementing these functions as methods on the struct. That way the struct has all the things and we pass less around. Will open an issue to track this.

bgins: Opened an issue for it: https://github.com/Lilypad-Tech/lilypad/issues/504

pkg/solver/matcher/matcher.go
narbs91: Beauty, sounds good!

narbs91 (approved):
LGTM!  Was able to run a cow-say locally and thank you for writing up those issues for tracking 🚀

PR #498, opened 2025-01-26
feat: Add Job Creator support endpoints to the Solver

kelindi:
This pull request makes the following changes to support the Job Creation Flow and API:

- [x] Adds the following routes:
`/job_offers/{id}`
Lets you fetch jobOffer information and deal status with a Job Offer ID
`/job_offers/{id}/results`
Lets you fetch the results for a job with a Job Offer ID

### DIFF
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -79,7 +79,7 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 	subrouter.HandleFunc("/job_offers", http.PostHandler(solverServer.addJobOffer)).Methods("POST")
 
 	subrouter.HandleFunc("/job_offers/{id}", http.GetHandler(solverServer.getJobOffer)).Methods("GET")
-	subrouter.HandleFunc("/job_offers/{id}/files", solverServer.getJobOfferFiles).Methods("GET")
+	subrouter.HandleFunc("/job_offers/{id}/files", solverServer.jobOfferDownloadFiles).Methods("GET")
 
 	subrouter.HandleFunc("/resource_offers", http.GetHandler(solverServer.getResourceOffers)).Methods("GET")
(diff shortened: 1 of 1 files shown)


pkg/solver/server.go
```diff
@@ -78,6 +78,9 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 	subrouter.HandleFunc("/job_offers", http.GetHandler(solverServer.getJobOffers)).Methods("GET")
 	subrouter.HandleFunc("/job_offers", http.PostHandler(solverServer.addJobOffer)).Methods("POST")
 
+	subrouter.HandleFunc("/job_offers/{id}", http.GetHandler(solverServer.getJobOffer)).Methods("GET")
+	subrouter.HandleFunc("/job_offers/{id}/results", solverServer.getJobOfferResult).Methods("GET")
```
bgins: What would you think about renaming this endpoint to `/job_offers/{id}/files`?

The other route that is used for downloading job outputs is `/deals/{id}/files`, so `files` would be consistent with that one. "results" is overloaded because we use it for retrieving results metadata in `/deals/{id}/result`.

pkg/solver/server.go
```diff
@@ -270,6 +284,111 @@ func (solverServer *solverServer) getResult(res corehttp.ResponseWriter, req *co
...
+
+		return nil
+	}()
+
+	if err != nil {
+		log.Ctx(req.Context()).Error().Msgf("error for route: %s", err.Error())
+		corehttp.Error(res, err.Error(), err.StatusCode)
+	}
+}
+
```
bgins: Could we make a helper function to handle some of this? Looks like a lot of the code is shared with `downloadFiles`.

bgins (approved):
Looks good! 🎉

PR #500, opened 2025-01-28
fix: silence usage text

### LINKED ISSUE #413: Silence help text on run errors
We currently print the help text on every command failure. Printing usage information is great when an error occurs parsing CLI options. However, it's noisy and unhelpful for errors reported afterwards.

Let's silence the help text after we have parsed CLI options. For example, in the `run` subcommand:

https://github.com/Lilypad-Tech/lilypad/blob/b467d0fa36f9720038bf5714d7ce1409a375d2a8/cmd/lilypad/run.go#L33-L37

we could set `cmd.SilenceUsage` to `true` after handling any `ProcessJobCreatorOptions` errors.

The silenced error output can be tested by temporarily returning an error from the first line of the `runJob` function:

https://github.com/Lilypad-Tech/lilypad/blob/b467d0fa36f9720038bf5714d7ce1409a375d2a8/cmd/lilypad/run.go#L46-L47

jdelamare:
### Summary

- [x] Silence usage text in solver, mediator, pow signal, resource provider, run commands

This PR silences usage text after command line options have been parsed.

### Task/Issue reference

Closes: #413

### Test plan

Run a command with a bad CLI option, usage should show. Add an intentional error after command line option parsing to see that usage is not shown.

### DIFF
--- a/cmd/lilypad/mediator.go
+++ b/cmd/lilypad/mediator.go
@@ -20,12 +20,13 @@ func newMediatorCmd() *cobra.Command {
 		Long:    "Start the lilypad mediator service.",
 		Example: "",
 		RunE: func(cmd *cobra.Command, _ []string) error {
-
 			network, _ := cmd.Flags().GetString("network")
 			options, err := optionsfactory.ProcessMediatorOptions(options, network)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


bgins (approved):
Looks good! Thanks for the contribution ❤️

PR #501, opened 2025-01-28
feat: Remove on chain deal agreement requirement

kelindi:
Makes the following changes to support the API Flow + in preparation for mainnet smart contracts and job creation flow
1. removes deal agreement requirements from the job creator side (still needs to be removed from the cli and resource provider)
2. set deal state to agreed on match (the states need to be revised and rearchitected to match the new smart contracts and job creation flow)

Supporting changes:

**Solver**
- Adds update JobStates helper function to update states across JobOffers, ResourceOffers, and Deal objects in the database in liu of smart contract updates
- Adds support to parse empty skeleton deal acceptance transactions from the JobCreator CLI

**Lilypad CLI**
- removes smart contract call to accept results

**Resource Provider**
- removes smart contract call to post results. Without this with the new solver changes the resource provider does not update its resource offer until it reaches the mandatory condition in the control loop (every 10 minutes).

 NOTE: all smart contract call removal is in preparation for integration with the new smart contracts for mainnet.

### DIFF
--- a/pkg/jobcreator/controller.go
+++ b/pkg/jobcreator/controller.go
@@ -387,19 +387,22 @@ func (controller *JobCreatorController) downloadResult(dealContainer data.DealCo
 
 func (controller *JobCreatorController) acceptResult(deal data.DealContainer) error {
 	controller.log.Debug("Accepting results for job", deal.ID)
-	txHash, err := controller.web3SDK.AcceptResult(deal.ID)
-	if err != nil {
-		return fmt.Errorf("error calling accept result tx for deal: %s", err.Error())
-	}
-	controller.log.Debug("accept result tx", txHash)
+	// txHash, err := controller.web3SDK.AcceptResult(deal.ID)
+	// if err != nil {
+	// 	return fmt.Errorf("error calling accept result tx for deal: %s", err.Error())
+	// }
+	// controller.log.Debug("accept result tx", txHash)
+	txHash := "0x"
 
 	// we have agreed to the deal so we need to update the tx in the solver
-	_, err = controller.solverClient.UpdateTransactionsJobCreator(deal.ID, data.DealTransactionsJobCreator{
+	_, err := controller.solverClient.UpdateTransactionsJobCreator(deal.ID, data.DealTransactionsJobCreator{
 		AcceptResult: txHash,
 	})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/solver/server.go
```diff
@@ -719,3 +731,36 @@ func (solverServer *solverServer) getValidationToken(res corehttp.ResponseWriter
 	// Respond with the JWT
 	return &http.ValidationToken{JWT: tokenString}, nil
 }
+
+
+
+func (solverServer *solverServer) updateJobstates(dealID string, state string) (error) {
```
bgins: nit: Could we rename this to `updateJobStates`? (capitalize the "S" in states)

pkg/jobcreator/controller.go
```diff
@@ -387,19 +387,22 @@ func (controller *JobCreatorController) downloadResult(dealContainer data.DealCo
...
-	txHash, err := controller.web3SDK.AcceptResult(deal.ID)
-	if err != nil {
-		return fmt.Errorf("error calling accept result tx for deal: %s", err.Error())
-	}
-	controller.log.Debug("accept result tx", txHash)
+	// txHash, err := controller.web3SDK.AcceptResult(deal.ID)
+	// if err != nil {
+	// 	return fmt.Errorf("error calling accept result tx for deal: %s", err.Error())
+	// }
+	// controller.log.Debug("accept result tx", txHash)
```
bgins: What would you think about removing the commented out code and instead leaving a comment stating what we did? Also could make it a TODO comment and suggest future work to fully remove this chain interaction.

pkg/resourceprovider/controller.go
```diff
@@ -526,12 +526,13 @@ func (controller *ResourceProviderController) runJob(ctx context.Context, deal d
...
-		createdResult.ID,
-		createdResult.DataID,
-		result.InstructionCount,
-	)
+	// txHash, err := controller.web3SDK.AddResult(
+	// 	deal.Deal.ID,
+	// 	createdResult.ID,
+	// 	createdResult.DataID,
+	// 	result.InstructionCount,
+	// )
```
bgins: Same note as above about commented out code.

bgins (approved):
Looks good! ✨

PR #507, opened 2025-01-31
feat: solver rate limiter exemption list

kelindi:

- [x] Adds a new env variable to add exemptions for the ratelimitter


pkg/solver/server.go
```diff
@@ -62,17 +63,44 @@ func NewSolverServer(
  *
 */
 
+func exemptIPKeyFunc(exemptIPs []string) func(r *corehttp.Request) (string, error) {
+	return func(r *corehttp.Request) (string, error) {
+		ip, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			ip = r.RemoteAddr
+		}
+
```
bgins: nit: Could we remove the extra newline here?

pkg/solver/server.go
```diff
@@ -62,17 +63,44 @@ func NewSolverServer(
...
+		for _, exemptIP := range exemptIPs {
+			if ip == exemptIP {
+				return "", nil
+			}
+		}
+		
+		return ip, nil
+	}
+}
+
```
bgins: Could we move this helper down below `disconnectCB` to keep it with the the other helpers?

pkg/solver/server.go
```diff
@@ -62,17 +63,44 @@ func NewSolverServer(
...
 	subrouter := router.PathPrefix(http.API_SUB_PATH).Subrouter()
 
 	subrouter.Use(http.CorsMiddleware)
 	subrouter.Use(otelmux.Middleware("solver", otelmux.WithTracerProvider(tracerProvider)))
+
+
+	exemptIPs := solverServer.options.RateLimiter.ExemptedIPs
+
+	log.Debug().Strs("exemptIPs", exemptIPs).Msg("Loaded rate limit exemptions")
+
```
bgins: What would you think about passing `solverServer.options.RateLimiter.ExemptedIPs` directly to `exemptIPKeyFunc` and logging the exempt IPs in there? Would clean things up a bit here.

pkg/options/server.go
```diff
@@ -29,6 +29,7 @@ func GetDefaultRateLimiterOptions() http.RateLimiterOptions {
 	return http.RateLimiterOptions{
 		RequestLimit: GetDefaultServeOptionInt("SERVER_RATE_REQUEST_LIMIT", 5),
 		WindowLength: GetDefaultServeOptionInt("SERVER_RATE_WINDOW_LENGTH", 10),
+		ExemptedIPs:  GetDefaultServeOptionStringArray("SERVER_RATE_EXEMPTED_IPS", []string{}),
```
bgins, line 32: Could we add a CLI option for this as well? There's an example in the PR where we first added the rate limiter options: https://github.com/Lilypad-Tech/lilypad/pull/419/commits/d65aa0d0b0f3b9cb2e0471c7ecd9315357739fc9.

pkg/solver/server.go
```diff
@@ -62,17 +63,44 @@ func NewSolverServer(
  *
 */
 
+func exemptIPKeyFunc(exemptIPs []string) func(r *corehttp.Request) (string, error) {
+	return func(r *corehttp.Request) (string, error) {
+		ip, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			ip = r.RemoteAddr
```
bgins: Using `r.RemoteAddr` won't work because we proxy through a Cloudflare tunnel. We need to determine the real IP address or we end up rate limiting based on the proxy IP address.

The `httprate` implementation of `KeyByRealIP` handles this correctly: https://github.com/go-chi/httprate/blob/9d627fb3c8a009b20ae198abe84b3a1d10a22a94/httprate.go#L52-L74. (Cloudflare adds an extra header we can key on.)

There's also a `canonicalizeIP` call in there. Haven't looked at it, but there may be things we want in it.

bgins (approved):
Looks good! 🎉

PR #508, opened 2025-02-05
fix: Use httprate.WithErrorHandler to exempt IPs

bgins:
### Summary

- [x] Update rate limiter to use errors instead of empty strings to indicate exemption

### Test plan

In the current state, the current `TestRateLimiter` integration test will fail because we are exempting localhost. Let's add a test for the exemption case and keep our existing test checking rate limiting.


kelindi (approved):
🚀

PR #510, opened 2025-02-10
feat: Api support + updates to job creation flow 

kelindi:
Makes the following changes to support the API Flow + in preparation for mainnet smart contracts and job creation flow

- removes deal agreement requirements from the job creation flow
- Adds the following routes to the solver:
`/job_offers/{id}`
Lets you fetch jobOffer information and deal status with a Job Offer ID
`/job_offers/{id}/results`
Lets you fetch the results for a job with a Job Offer ID


pkg/http/utils.go
```diff
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	stdlog "log"
+	"net"
 	"net/http"
 	"net/url"
```
narbs91, line 14: If you're bringing in the entire `net` package we could probs get rid of these or conversely you can change the `net` import on line 12 to just bring in the package you need

pkg/solver/server.go
```diff
@@ -51,33 +51,49 @@ func NewSolverServer(
...
 	router := mux.NewRouter()
 
 	subrouter := router.PathPrefix(http.API_SUB_PATH).Subrouter()
 
 	subrouter.Use(http.CorsMiddleware)
 	subrouter.Use(otelmux.Middleware("solver", otelmux.WithTracerProvider(tracerProvider)))
+
+	exemptIPs := solverServer.options.RateLimiter.ExemptedIPs
+
+	log.Debug().Strs("exemptIPs", exemptIPs).Msg("Loaded rate limit exemptions")
```
bgins: nit: Maybe move this below where we apply the rate limiting (around line 90)? It would confirm for us that we have applied the exempted IPs.

pkg/solver/controller.go
```diff
@@ -448,8 +448,12 @@ func (controller *SolverController) addDeal(ctx context.Context, deal data.Deal)
 
 	controller.log.Info("add deal", deal)
 
+	//creates deal container and sets state to agreed
+	dealContainer := data.GetDealContainer(deal)
+	dealContainer.State = data.GetAgreementStateIndex("DealAgreed")
```
narbs91, line 453: [FYI] we'll need to change this to `DealCreated` in when the new contracts are out

pkg/solver/server.go
narbs91: This should probs be at the `info` level

pkg/solver/server.go
```diff
@@ -660,3 +763,36 @@ func (solverServer *solverServer) getValidationToken(res corehttp.ResponseWriter
...
+	return err
+   }
+
+   // update the job offer state
+   _, err = solverServer.controller.updateJobOfferState(deal.Deal.JobOffer.ID, deal.ID, data.GetAgreementStateIndex(state))
+   if err != nil {
+	return err
+   }
+
+   // update the resource offer state
```
narbs91: [nit] Spacing here is off here

pkg/solver/server.go
```diff
@@ -625,6 +622,112 @@ func (solverServer *solverServer) uploadFiles(res corehttp.ResponseWriter, req *
 	}
 }
 
+func (solverServer *solverServer) jobOfferDownloadFiles(res corehttp.ResponseWriter, req *corehttp.Request) {
```
narbs91, line 650: [nit] Maybe we should change the name of this function to be getResultsForJob?

pkg/solver/server.go
```diff
@@ -346,7 +392,18 @@ func (solverServer *solverServer) addResult(results data.Result, res corehttp.Re
...
 	}
 	results.DealID = id
-	return solverServer.store.AddResult(results)
+
+	storedResult, err := solverServer.store.AddResult(results)
+	if err != nil {
+		return nil, err
+	}
+
+	err = solverServer.updateJobStates(id, "ResultsSubmitted")
```
narbs91, line 426: Probs better if we declare all these states as constants at the top of the file

pkg/solver/server.go
```diff
@@ -177,6 +193,25 @@ func (solverServer *solverServer) disconnectCB(connParams http.WSConnectionParam
...
+		ip, err := httprate.KeyByRealIP(r)
+		if err != nil {
+			log.Error().Err(err).Msg("error getting real ip")
+			return ip, err
+		}
+
+		// Check if the IP is in the exempt list
+		for _, exemptIP := range exemptIPs {
+			if http.CanonicalizeIP(exemptIP) == ip {
+				return "", fmt.Errorf("RATE_LIMIT_EXEMPT")
```
narbs91: Is there a reason you're returning this as an error instead of just the string or maybe even a boolean?  To me i'd think this function should say if the ip is exempt via a bool

pkg/solver/server.go
kelindi: Its closer to the exiting downloadFiles function under the deals route so I think the current naming makes more sense. The results endpoint we already have don't actually return the files.

pkg/solver/server.go
kelindi: The httprate.WithKeyFuncs() functon where this is used expects a function that returns (string, error)

pkg/solver/server.go
narbs91: Sure thing, all good

pkg/solver/server.go
narbs91: Ah gotcha, didn't realize that

pkg/http/utils.go
kelindi: we aren't able to import net/IpParse separately so the current implementation makes the most sense

.github/workflows/test.yml
```diff
@@ -53,7 +53,8 @@ jobs:
       - name: Start solver
         run: |
           ./stack solver --disable-telemetry=true --api-host="" > solver.log &
-          sleep 5
+          sleep 30
+          cat solver.log
```
bgins: Could we bring this back down. Perhaps 15 seconds will be enough? Also, let's remove the `cat solver.log`.

pkg/options/server.go
```diff
@@ -80,5 +85,12 @@ func CheckServerOptions(options http.ServerOptions) error {
 	if options.AccessControl.ValidationTokenKid == "" {
 		return fmt.Errorf("SERVER_VALIDATION_TOKEN_KID is required")
 	}
+	if len(options.RateLimiter.ExemptedIPs) > 0 {
+		for _, ip := range options.RateLimiter.ExemptedIPs {
+			if net.ParseIP(ip) == nil {
```
bgins, line 100: Awesome, thanks for adding! ❤️

bgins (approved):
Looks good. We could potentially skip the rate limiting tests so they show up as skipped in test reporting, but since we are following this up quickly let's go ahead.

Left one last comment and your signing key is showing unverified on the commits, so let's clear that up and get this merged!

PR #512, opened 2025-02-14
feat: Add global job timeout

### LINKED ISSUE #511: Add global job timeout
## General Description

Add a global job timeout to clear out stale jobs.

This functionality affects the solver, job creator and resource provider.

We would like to add a global job timeout to clear out stuck, stale, and abandoned jobs. The solver will check job offer creation times and mark a job as cancelled if it has outlived the global timeout.

This mechanism will be superseded by timeouts in module specs and job offers when we have implemented them.

bgins:
### Summary

- [x] Add expired job cancellation to the solver control loop
  - [x] Add `JOB_TIMEOUT_SECONDS` config
  - [x] Add `JobTimedOut` terminal agreement state
  - [x] Add cancelled job metrics 
  - [x] Add cancelled job traces
- [x] Add `Active` job offer query filter
- [x] Add `Cancelled` job offer query filter
  - [x] Remove `IncludeCancelled` filter and integrate it into `Cancelled` 

We have also made a couple of related changes and small cleanups:

- [x] Rename deal stats metrics to job stats metrics
- [x] Record job stat metrics for cancelled job offers that are not involved in a deal
- [x] Clean up allowed resource providers between test runs. We missed it while rebasing #481.
- [x] Update comments in solver `solve` function
- [x] Add `BoolPointer` language helper to more easily construct pointers to bools

The primary goal of this pull request is adding a global job timeout for jobs that are stuck, stale, or abandoned. The global timeout will be superseded by timeouts in module specs and job offers when we have implemented them.

We have also added new `Active` and `Cancelled` job offer query filters in support of the global timeout.

### Task/Issue reference

Closes: #511

### Test plan

Start the base services. Start the solver with the job offer timeout set to ten seconds:

```
./stack solver --job-timeout-seconds 10
```

Submit a job without a resource provider running. The job should fail and report a job timed out error message.

Start a resource provider and run another job. Again, the job should be cancelled, but this time in flight. The job creator should see the error message and the resource provider should repost their resource offer.

See the details below if you would like to confirm the activity through metrics and traces.

### Details

**`Active` and `Cancelled` job offer filter queries**

We have added two new job offer filter queries. 

When `Active` is `true`, only jobs in `DealNegotiating`, `DealAgreed`, or `ResultsSubmitted` states are included.

Query active job offers from the solver server with the `active` querstring param. In local development:

```sh
curl http://localhost:8081/api/v1/job_offers?active=true
```

The `Cancelled` query filter is a `*bool` where:

- `nil` includes all jobs
- `true` includes jobs in `JobOfferCancelled` and `JobTimedOut` states
- `false` includes jobs _not_ in a `JobOfferCancelled` or `JobTimedOut` state

We removed the `IncludeCancelled` filter query which is now performed when `Cancelled` is `nil`.

Query jobs on the solver server using the `cancelled` querystring param. In local development:

```sh
curl http://localhost:8081/api/v1/job_offers                    # all jobs
curl http://localhost:8081/api/v1/job_offers?cancelled=true     # cancelled jobs   
curl http://localhost:8081/api/v1/job_offers?cancelled=false    # not cancelled jobs
```

**`JobTimedOut` agreement state**

We have added a new `JobTimedOut` agreement state to describe jobs that were cancelled because they expired. 

We considered using the existing `JobOfferCancelled` state, but it is not coherent when we want to mark resource offers and deals with an expired state. In addition, our mechanism for reporting what happened to the job creator is limited, so `JobTimedOut` serves as a tag to help us log a good error message for the job creator.

When a job times out before a deal was made, the job offer is marked with the `JobTimedOut` state. When a deal has already been made, the job offer, resource offer, and deal are all marked with the `JobTimedOut` state.

If a resource provider attempts to upload files or report results for an expired job, the solver server rejects their post with an error reporting the job as timed out.

**`JOB_TIMEOUT_SECONDS` config**

We have added a new config to set the global timeout in seconds. The default value is ten minutes.

Set the job timeout with an environment variable of CLI option. For example, to expire jobs after 30 seconds:

```
./stack solver --job-timeout-seconds 30
JOB_TIMEOUT_SECONDS=30 ./stack solver
```

**Observability**

The deal metrics have been updated to reflect overall job metrics. We want to include metrics about job offers that were never in a deal, and we record these from jobs that expired or were cancelled.

The metrics are reported in an updated `Solver Metrics` dashboard:

timed-out-metrics

The updated dashboard is available in this observability pull request: https://github.com/Lilypad-Tech/observability/pull/27

Metrics are not enabled by default in local development, but they can be turned on with:

```sh
export ENABLE_METRICS=true
export METRICS_URL="http://localhost:8500"
export METRICS_TOKEN="eyJhbGciOiJFZERTQSIsInR5cCI6IkpXVCJ9.eyJhdXRob3JpemVkIjp0cnVlLCJ1c2VyIjoicmVzb3VyY2UtcHJvdmlkZXIifQ.n36M_ngwC4XPQ_pEkkWAnPiOinnx6-0VO1v_WgCTUEERD7b_p9KHCU6SY5bUdFh5UXRZHAhc1gfyc7rjAnmeDQ"
  ```

We have also included traces over our new `cancelExpiredJobs` solver controller function. Traces are also disabled by default in local development, but can be enabled with the `--disable-telemetry=false` CLI option.

The traces can be queried by searching TraceQL with the `cancel_expired_jobs` span name and a non-empty expired jobs span attribute:

```
{name="cancel_expired_jobs" && span.expired_job_offers!="[]"}
```

When a job is cancelled before a match, the trace will look like:

cancelled-before-match-trace

If a job is cancelled after a match, it will also report a deal ID:

cancelled-post-match

### Related issues or PRs

Observability dashboard PR: https://github.com/Lilypad-Tech/observability/pull/27

### DIFF
--- a/pkg/options/solver.go
+++ b/pkg/options/solver.go
@@ -10,14 +10,13 @@ import (
 
 func NewSolverOptions() solver.SolverOptions {
 	options := solver.SolverOptions{
-		Server:    GetDefaultServerOptions(),
-		Store:     GetDefaultStoreOptions(),
-		Web3:      GetDefaultWeb3Options(),
-		Services:  GetDefaultServicesOptions(),
-		Telemetry: GetDefaultTelemetryOptions(),
-		Metrics:   GetDefaultMetricsOptions(),
-		// Timeout in seconds with a default of ten minutes
-		JobTimeoutSeconds: GetDefaultServeOptionInt("JOB_TIMEOUT_SECONDS", 600),
+		Server:            GetDefaultServerOptions(),
+		Store:             GetDefaultStoreOptions(),
+		Web3:              GetDefaultWeb3Options(),
+		Services:          GetDefaultServicesOptions(),
+		Telemetry:         GetDefaultTelemetryOptions(),
+		Metrics:           GetDefaultMetricsOptions(),
+		JobTimeoutSeconds: GetDefaultServeOptionInt("JOB_TIMEOUT_SECONDS", 600), // 10 minutes
 	}
 	options.Web3.Service = system.SolverService
 	return options
--- a/pkg/options/solver.go
+++ b/pkg/options/solver.go
@@ -17,7 +17,7 @@ func NewSolverOptions() solver.SolverOptions {
 		Telemetry: GetDefaultTelemetryOptions(),
 		Metrics:   GetDefaultMetricsOptions(),
 		// Timeout in seconds with a default of ten minutes
-		JobOfferTimeout: GetDefaultServeOptionInt("JOB_OFFER_TIMEOUT", 600),
+		JobTimeoutSeconds: GetDefaultServeOptionInt("JOB_TIMEOUT_SECONDS", 600),
 	}
 	options.Web3.Service = system.SolverService
 	return options
@@ -31,8 +31,8 @@ func AddSolverCliFlags(cmd *cobra.Command, options *solver.SolverOptions) {
 	AddTelemetryCliFlags(cmd, &options.Telemetry)
 	AddMetricsCliFlags(cmd, &options.Metrics)
 	cmd.PersistentFlags().IntVar(
-		&options.JobOfferTimeout, "job-offer-timeout", options.JobOfferTimeout,
-		`The global timeout for job offers in seconds (JOB_OFFER_TIMEOUT).`,
+		&options.JobTimeoutSeconds, "job-timeout-seconds", options.JobTimeoutSeconds,
+		`The global timeout for jobs in seconds (JOB_TIMEOUT_SECONDS)`,
 	)
 }
 
@@ -57,8 +57,8 @@ func CheckSolverOptions(options solver.SolverOptions) error {
 	if err != nil {
 		return err
 	}
-	if options.JobOfferTimeout <= 0 {
-		return fmt.Errorf("JOB_OFFER_TIMEOUT must be greater than zero")
+	if options.JobTimeoutSeconds <= 0 {
+		return fmt.Errorf("JOB_TIMEOUT_SECONDS must be greater than zero")
 	}
 	return nil
 }
--- a/pkg/options/solver.go
+++ b/pkg/options/solver.go
@@ -1,6 +1,8 @@
 package options
 
 import (
+	"fmt"
+
 	"github.com/lilypad-tech/lilypad/pkg/solver"
 	"github.com/lilypad-tech/lilypad/pkg/system"
 	"github.com/spf13/cobra"
@@ -8,12 +10,13 @@ import (
 
 func NewSolverOptions() solver.SolverOptions {
 	options := solver.SolverOptions{
-		Server:    GetDefaultServerOptions(),
-		Store:     GetDefaultStoreOptions(),
-		Web3:      GetDefaultWeb3Options(),
-		Services:  GetDefaultServicesOptions(),
-		Telemetry: GetDefaultTelemetryOptions(),
-		Metrics:   GetDefaultMetricsOptions(),
+		Server:          GetDefaultServerOptions(),
+		Store:           GetDefaultStoreOptions(),
+		Web3:            GetDefaultWeb3Options(),
+		Services:        GetDefaultServicesOptions(),
+		Telemetry:       GetDefaultTelemetryOptions(),
+		Metrics:         GetDefaultMetricsOptions(),
+		JobOfferTimeout: GetDefaultServeOptionInt("JOB_OFFER_TIMEOUT", 600),
 	}
 	options.Web3.Service = system.SolverService
 	return options
@@ -26,6 +29,10 @@ func AddSolverCliFlags(cmd *cobra.Command, options *solver.SolverOptions) {
 	AddServicesCliFlags(cmd, &options.Services)
 	AddTelemetryCliFlags(cmd, &options.Telemetry)
 	AddMetricsCliFlags(cmd, &options.Metrics)
+	cmd.PersistentFlags().IntVar(
+		&options.JobOfferTimeout, "job-offer-timeout", options.JobOfferTimeout,
+		`The global timeout for job offers in seconds (JOB_OFFER_TIMEOUT).`,
+	)
 }
 
 func CheckSolverOptions(options solver.SolverOptions) error {
@@ -48,6 +55,9 @@ func CheckSolverOptions(options solver.SolverOptions) error {
 	err = CheckMetricsOptions(options.Metrics)
 	if err != nil {
 		return err
+	}
+	if options.JobOfferTimeout <= 0 {
+		return fmt.Errorf("JOB_OFFER_TIMEOUT must be greater than zero")
 	}
 	return nil
 }
--- a/pkg/options/solver.go
+++ b/pkg/options/solver.go
@@ -10,12 +10,13 @@ import (
 
 func NewSolverOptions() solver.SolverOptions {
 	options := solver.SolverOptions{
-		Server:          GetDefaultServerOptions(),
-		Store:           GetDefaultStoreOptions(),
-		Web3:            GetDefaultWeb3Options(),
-		Services:        GetDefaultServicesOptions(),
-		Telemetry:       GetDefaultTelemetryOptions(),
-		Metrics:         GetDefaultMetricsOptions(),
+		Server:    GetDefaultServerOptions(),
+		Store:     GetDefaultStoreOptions(),
+		Web3:      GetDefaultWeb3Options(),
+		Services:  GetDefaultServicesOptions(),
+		Telemetry: GetDefaultTelemetryOptions(),
+		Metrics:   GetDefaultMetricsOptions(),
+		// Timeout in seconds with a default of ten minutes
 		JobOfferTimeout: GetDefaultServeOptionInt("JOB_OFFER_TIMEOUT", 600),
 	}
 	options.Web3.Service = system.SolverService
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -282,6 +282,14 @@ func (controller *SolverController) solve(ctx context.Context) error {
 	ctx, span := controller.tracer.Start(ctx, "solve")
 	defer span.End()
 
+	// Remove expired job offers
+	err := controller.cancelExpiredJobs(ctx)
+	if err != nil {
+		span.SetStatus(codes.Error, "remove expired offers failed")
+		span.RecordError(err)
+		return err
+	}
+
 	// find out which deals we can make from matching the offers
 	deals, err := matcher.GetMatchingDeals(ctx, controller.store, controller.updateJobOfferState, controller.log, controller.tracer, controller.meter)
 	if err != nil {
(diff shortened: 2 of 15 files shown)

### CODE REVIEW

File: pkg/options/solver.go [hunk 1]
```diff
@@ -26,6 +29,10 @@ func AddSolverCliFlags(cmd *cobra.Command, options *solver.SolverOptions) {
 	AddServicesCliFlags(cmd, &options.Services)
 	AddTelemetryCliFlags(cmd, &options.Telemetry)
 	AddMetricsCliFlags(cmd, &options.Metrics)
+	cmd.PersistentFlags().IntVar(
+		&options.JobOfferTimeout, "job-offer-timeout", options.JobOfferTimeout,
```
narbs91: Can we change `job-offer-timeout` to `job-offer-timeout-in-seconds` so that it's clear what unit of measure should be set?

pkg/options/solver.go
```diff
@@ -49,6 +56,9 @@ func CheckSolverOptions(options solver.SolverOptions) error {
 	if err != nil {
 		return err
 	}
+	if options.JobOfferTimeout <= 0 {
```
narbs91: Did we want to enforce a minimum timeout at this point if someone sets it?  Like someone could technically set a timeout of a second if they wanted to (but obviously probably wouldn't).  Maybe jus the 600s you put in the default?

pkg/options/solver.go
```diff
@@ -1,19 +1,22 @@
...
-		Services:  GetDefaultServicesOptions(),
-		Telemetry: GetDefaultTelemetryOptions(),
-		Metrics:   GetDefaultMetricsOptions(),
+		Server:          GetDefaultServerOptions(),
+		Store:           GetDefaultStoreOptions(),
+		Web3:            GetDefaultWeb3Options(),
+		Services:        GetDefaultServicesOptions(),
+		Telemetry:       GetDefaultTelemetryOptions(),
+		Metrics:         GetDefaultMetricsOptions(),
+		JobOfferTimeout: GetDefaultServeOptionInt("JOB_OFFER_TIMEOUT", 600),
```
narbs91: [nit] Maybe have the 600 as a constant variable so that it can be easily set/referenced if needed?

narbs91: Maybe we can also rename this `JobOfferTimeoutInSeconds`

pkg/solver/controller.go
```diff
@@ -321,6 +336,67 @@ func (controller *SolverController) solve(ctx context.Context) error {
...
+		now := time.Now().UnixMilli()
+		if now-int64(jobOffer.JobOffer.CreatedAt) > int64(controller.options.JobOfferTimeout*1000) {
+			if jobOffer.DealID == "" {
+				// Cancel expired job offers
+				_, err := controller.updateJobOfferState(jobOffer.ID, jobOffer.DealID, data.GetAgreementStateIndex("JobTimedOut"))
+				if err != nil {
+					controller.log.Error("update expired job offer state failed", err)
+					span.SetStatus(codes.Error, "update expired job offer state failed")
+					span.RecordError(err)
+					return err
```
narbs91: Do we completely want to break the loop on failure vs continuing?

pkg/solver/controller.go
```diff
@@ -321,6 +336,67 @@ func (controller *SolverController) solve(ctx context.Context) error {
...
+				}
+			} else {
+				// Cancel expired job offers, resource offers, and deals
+				_, err := controller.updateDealState(jobOffer.DealID, data.GetAgreementStateIndex("JobTimedOut"))
+				if err != nil {
+					controller.log.Error("update expired deal state failed", err)
+					span.SetStatus(codes.Error, "update expired deal state failed")
+					span.RecordError(err)
+					return err
```
narbs91: Ditto here?

pkg/solver/solver.go
```diff
@@ -15,12 +15,13 @@ import (
...
-	Services  data.ServiceConfig
-	Telemetry system.TelemetryOptions
-	Metrics   system.MetricsOptions
+	Server          http.ServerOptions
+	Store           store.StoreOptions
+	Web3            web3.Web3Options
+	Services        data.ServiceConfig
+	Telemetry       system.TelemetryOptions
+	Metrics         system.MetricsOptions
+	JobOfferTimeout int
```
narbs91: Same comment here about changing to `JobOfferTimeoutInSeconds`

pkg/solver/store/db/db.go
```diff
@@ -159,8 +159,23 @@ func (store *SolverStoreDatabase) GetJobOffers(query store.GetJobOffersQuery) ([
...
+			data.GetAgreementStateIndex("DealAgreed"),
+			data.GetAgreementStateIndex("ResultsSubmitted"),
+		})
+	}
+	if query.Cancelled {
+		q = q.Where("state IN (?)", []uint8{
+			data.GetAgreementStateIndex("JobOfferCancelled"),
+			data.GetAgreementStateIndex("JobTimedOut"),
+		})
+	} else if !query.IncludeCancelled {
```
narbs91: Does there need to be both a Cancelled and a IncludeCancelled?  Is there a different purpose between the two?

pkg/solver/store/store.go
```diff
@@ -20,6 +20,13 @@ type GetJobOffersQuery struct {
 	// we use the DealID property of the jobOfferContainer to tell if it's been matched
 	NotMatched bool `json:"not_matched"`
 
+	// Active job offers are umatched or in an in-progress deal.
+	// This includes the DealNegotiating, DealAgreed, or ResultsSubmitted states.
+	Active bool `json:"in_progress"`
+
+	// Cancelled job offers are in a JobOfferCancelled or JobTimedOut state
+	Cancelled bool `json:"cancelled"`
```
narbs91: Ah I see, ones for including results the other is for clearing out the stale offers.  Perhaps we can change the name of this to `inactive` or `terminated` so that it doesn't clash with `IncludeCancelled`?

pkg/options/solver.go
bgins: It's a bit verbose. The help text for the option says it:

https://github.com/Lilypad-Tech/lilypad/blob/bcf3160d901e52869cc95f0ccf6e0e31d8a30925/pkg/options/solver.go#L34

That seem like enough to communicate it?

pkg/options/solver.go
bgins: There are legitimate use cases for a very low timeout. For example, testing this PR we want to set a low timeout for testing. Can see how it could be useful for other local development scenarios and future integration tests.

pkg/options/solver.go
bgins: Would be open to this change if we have a specific use case in mind. Could see readability as a potential motivation, but it would be an exception to how we generally set the defaults for configs.

Perhaps we could take it up separately and think about what patterns we want for our options?

bgins: See above, we document in the help text, which may be enough? Perhaps we add a comment here to make it clearer when reading the code.

pkg/solver/solver.go
bgins: Same comment as above 🙂

pkg/solver/store/db/db.go
bgins: Yes, the purpose of `IncludeCancelled` is when you want to query all offers _except_ cancelled ones. The new `Cancelled` query is when you want to retrieve _only_ cancelled offers.

bgins: Oh actually, maybe we can repurpose `Cancelled`:

- When `Cancelled` is `true`, only return offers in a `JobOfferCancelled` or a `JobTimedOut` state.
- When `Cancelled` is `false`, return all offers not in a `JobOfferCancelled` or a `JobTimedOut` state.

pkg/solver/store/store.go
bgins: Wrote some more thoughts about this above, we may be able to simplify it to only use `Cancelled`.

pkg/solver/controller.go
bgins: Yeah great point! Logging and tracing should be enough here. Will update. 👌

bgins: Done ✅

pkg/solver/controller.go
bgins: Removed ✅

pkg/options/solver.go
narbs91: Ah good point, makes sense to me

pkg/options/solver.go
narbs91: i personally always like to be precise with this types of things but don't mind if we want to just leave the text to be the reference.  Will defer to you

pkg/options/solver.go
narbs91: The use case I was thinking of was if we wanted to enforce this a minimum timeout to set but if we're not doing that, ok to leave it as is

narbs91: Yup, comment would do the trick too

bgins: Sounds good! Added a comment. 👌

pkg/options/solver.go
bgins: We discussed this a bit more and decided that there are places where we don't have the CLI help text to provide context. We've updated this to `JOB_TIMEOUT_SECONDS` to add the unit and to communicate that we are expiring not just offers, but jobs.

pkg/solver/server.go
```diff
@@ -242,9 +243,17 @@ func (solverServer *solverServer) getJobOffers(res corehttp.ResponseWriter, req
 	if notMatched := req.URL.Query().Get("not_matched"); notMatched == "true" {
 		query.NotMatched = true
 	}
-	if includeCancelled := req.URL.Query().Get("include_cancelled"); includeCancelled == "true" {
-		query.IncludeCancelled = true
+	if active := req.URL.Query().Get("active"); active == "true" {
+		query.Active = true
 	}
+	if cancelled := req.URL.Query().Get("cancelled"); cancelled != "" {
+		if val, err := strconv.ParseBool(cancelled); err == nil {
```
narbs91, line 250: Is this suppose to go into the if block if `err==nil`?

pkg/solver/store/memory/store.go
```diff
@@ -105,9 +105,26 @@ func (s *SolverStoreMemory) GetJobOffers(query store.GetJobOffersQuery) ([]data.
...
 		}
-		if !query.IncludeCancelled && jobOffer.State == data.GetAgreementStateIndex("JobOfferCancelled") {
+		if query.Active &&
+			jobOffer.State != data.GetAgreementStateIndex("DealNegotiating") &&
+			jobOffer.State != data.GetAgreementStateIndex("DealAgreed") &&
+			jobOffer.State != data.GetAgreementStateIndex("ResultsSubmitted") {
 			matching = false
 		}
+		if query.Cancelled != nil {
+			isCancelled := jobOffer.State == data.GetAgreementStateIndex("JobOfferCancelled") ||
```
narbs91, line 115: [nit] maybe change this variable to be `isJobOfferInStaleState` so that it doesn't clash with the `query.Cancelled`

narbs91 (approved):
LGTM!

pkg/solver/store/db/db.go
bgins: Following up, we added a `nil` state where `Cancelled` is not present to query for all job offers.

pkg/solver/store/memory/store.go
bgins: Updated to improve readability ✅

PR #513, opened 2025-02-14
feat: Anura Rate Limiter

kelindi:
### Summary

- [x] Adds new separate routes for Anura
- [x] Creates a Rate Limitting middleware that only allows Anura to interface with these endpoints providing a valid signature


pkg/options/server.go
```diff
@@ -24,14 +24,14 @@ func GetDefaultAccessControlOptions() http.AccessControlOptions {
 		ValidationTokenSecret:           GetDefaultServeOptionString("SERVER_VALIDATION_TOKEN_SECRET", ""),
 		ValidationTokenExpiration:       GetDefaultServeOptionInt("SERVER_VALIDATION_TOKEN_EXPIRATION", 604800), // one week
 		ValidationTokenKid:              GetDefaultServeOptionString("SERVER_VALIDATION_TOKEN_KID", ""),
+		AnuraAddresses:                 GetDefaultServeOptionStringArray("ANURA_ADDRESSES", []string{}),
```
bgins: nit: Minor formatting issue here, off by one space.

pkg/solver/anuraLimit_test.go
```diff
@@ -0,0 +1,156 @@
...
+	expectedLimit  int
+	expectedUnauth int
+}
+
+// This test suite sends 10 requests to Anura endpoints. We test both valid and invalid
+// signatures. Invalid signatures should be rate limited after 5 attempts.
+func TestAnuraRateLimiter(t *testing.T) {
+	paths := []string{
+		"/api/v1/anura/job_offers",
+	}
```
bgins, line 25: Would this test work over multiple paths? We have the machinery to test them in parallel, could be interesting for validating the unauthorized and limited cases.

pkg/solver/ratelimit_test.go
bgins, line 1: Could we do some renaming in this file to remove the notion of exemption?

https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/solver/ratelimit_test.go#L41-L59

bgins:
A few places where we have references to the exemption code that can be removed:

https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/http/types.go#L26
https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/options/server.go#L98-L104
https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/solver/server.go#L73-L91
https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/solver/server.go#L99
https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/solver/server.go#L237-L254

We can probably also get rid of this canonicalize helper we added for the exemption handler:

https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/http/utils.go#L506-L531

pkg/solver/ratelimit_test.go
bgins: Could we also rewrite this comment to remove the considerations around exemption?

https://github.com/Lilypad-Tech/lilypad/blob/0fd63044fd88df116b7dcb1cce703d4325b4db11/pkg/solver/ratelimit_test.go#L27-L33

pkg/solver/anuraLimit_test.go
kelindi: all the other endpoints require posting a job/testing the status of a job. So they rely on other functionality working. Thought it would be good to test the rate limiter independently from other components.

bgins: Yep makes sense! Do you think there will be a case in the future where we need to test multiple paths? If not, we could probably simplify the test by removing the machinery for parallel execution.

pkg/solver/anuraLimit_test.go
bgins, line 1: nit: Could we rename this to `anuralimit_test.go` (with a lowercase `l`). From what I've seen, camel case naming is not used in Go code bases.

pkg/solver/server.go
```diff
@@ -122,6 +100,37 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 
 	subrouter.HandleFunc("/validation_token", http.GetHandler(solverServer.getValidationToken)).Methods("GET")
 
+	//anura subrouter
+
+	failedAuthLimiter := httprate.Limit(
+		5,
+		60*time.Second,
+		httprate.WithKeyFuncs(httprate.KeyByRealIP),
+	)
```
bgins: Could we add configs for these values? Or possibly use the existing rate limiting configs?

bgins (approved):
Looks good! 🎉

PR #514, opened 2025-02-16
fix: RP error handling

walkah:
### Summary

This pull request makes the following changes to help resource provider reliability:

- Change when we run `ensureResourceOffers` to run after every `runJob` (whether it succeeds or fails)
- Catches non-"completed" terminal bacalhau execution states (failed and stopped) to properly exit the execution loop


narbs91 (approved):
LGTM!

PR #517, opened 2025-02-20
feat: Add job run stats reporting

### LINKED ISSUE #506: Add job run timing
## General Description

Add job run timing on the solver and report deals and timings to the stats API.

This functionality affects the solver and the stats API. It has an indirect impact on module authors and job creators who will use the data to establish job timeouts.

A job run starts when a job offer is created and ends when the job creator finishes downloading the job outputs. We will only time successful jobs at first and will consider timing failed jobs in future work.

bgins:
### Summary

- [x] Add solver `stats` package for reporting stats
  - [x] Add `stats` options
  - [x] Add `HTTPStats` implementation
  - [x] Add `NoopStats` implementation
- [x] Post job runs with job run timing to stats API
  - [x] Add `GetShortcut` helper to construct module identifier string
- [x] Post job completed (no validation) stat to stats API 

This pull request adds our initial stats reporting implementation to the solver. We start with reporting job runs with timing and jobs completed (no validation) to the stats API.

We have also updated the `downloadFiles` and `jobOfferDownloadFiles` handlers to improve error handling.

### Task/Issue reference

Closes: #506

### Test plan

Start the base services and the stats API. The stats API can re-use the postgres database from base services, so starting it should only require:

```sh
go run .
```

Start the solver enabling the stats API and providing the local stats URL:

```sh
```

Start the resource provider and run a job. Wait for the job to complete, then connect to the database. For example, using `psql`:

```bash
docker exec -it postgres psql -U postgres -d stats_api
```

Check the `job_runs` and `reputations` tables for records reflecting the job run:

```sql
select * from job_runs;
select * from reputations;
```

The `reputations` table should have a record with a `true` value in the `job_completed_no_validation` column.

### Details

**`stats` package**

The new stats package includes `HTTPStats` and `NoopStats` implementations. The `HTTPStats` implementation posts to the stats API. The `NoopStats` implementation is used when stats are disabled.

`HTTPStats` adds address and signature headers for authentication when reporting to the stats API.

**Reporting Job runs**

The `stats` package reports job runs including the following:

https://github.com/Lilypad-Tech/lilypad/blob/e303af4ab4df06df0490734d00b8aeb1440a8e11/pkg/solver/stats/jobrun.go#L15-L24

A job run starts when a job offer is created and ends when the job creator finishes downloading the job outputs. We only time successful runs in this pull request.

We rely on the `CreatedAt` field in the job offer as the starting point for a job. To avoid inaccurate reports, the solver only accepts job offers with recent creation times (see #523).

**Reporting reputation**

The `stats` reputation implementation includes a builder to collect multiple reputation stats before reporting. This pull request only reports jobs completed without validation, but in the future we will collect multiple reputation stats and report them in bulk where possible.

**Stats configs**

We have new `ENABLE_STATS` and `STATS_URL` configs. When `ENABLE_STATS` is `true`, `STATS_URL` must be set.

Enable stats with an environment variable of CLI option:

```
ENABLE_STATS=true ./stack solver
./stack solver --enable-stats true
```

Set the Stats API URL with an environment variable of CLI option:

```
STATS_URL=http://localhost:8080 ./stack solver
./stack solver --stats-url http://localhost:8080
```

**Download file updates**

We have also updated the `downloadFiles` and `jobOfferDownloadFiles` handlers to use our `http.GetHandler` middleware, which encodes JSON errors before sending them to the user. 

This change required us to return a `EmptyResponse` in success cases. The empty response should be ignored by downloaders because they are interested in the file download stream.

### Related issues or PRs

Epic: https://github.com/Lilypad-Tech/internal/issues/348

### DIFF
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -605,6 +605,7 @@ func (solverServer *solverServer) handleFileDownload(dirPath string, res corehtt
 	for _, file := range files {
 		info, err := file.Info()
 		if err != nil {
+			log.Warn().Msgf("expected file renamed or moved: %v", err)
 			continue
 		}
 		if info.Mode().IsRegular() {
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -577,6 +577,11 @@ func (solverServer *solverServer) downloadFiles(res corehttp.ResponseWriter, req
 
 	if err := solverServer.handleFileDownload(GetDealsFilePath(id), res, func() {
 		solverServer.stats.PostJobRun(solverServer.store, deal)
+		solverServer.stats.PostReputation(deal.ResourceProvider,
+			stats.NewReputationBuilder().
+				WithJobCompletedNoValidation(true).
+				Build(),
+		)
 	}); err != nil {
 		return EmptyResponse{}, err
 	}
@@ -777,6 +782,11 @@ func (solverServer *solverServer) jobOfferDownloadFiles(res corehttp.ResponseWri
 
 	if err := solverServer.handleFileDownload(GetDealsFilePath(jobOffer.DealID), res, func() {
 		solverServer.stats.PostJobRun(solverServer.store, deal)
+		solverServer.stats.PostReputation(deal.ResourceProvider,
+			stats.NewReputationBuilder().
+				WithJobCompletedNoValidation(true).
+				Build(),
+		)
 	}); err != nil {
 		return EmptyResponse{}, err
 	}
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -81,7 +81,7 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 	subrouter.HandleFunc("/job_offers", http.PostHandler(solverServer.addJobOffer)).Methods("POST")
 
 	subrouter.HandleFunc("/job_offers/{id}", http.GetHandler(solverServer.getJobOffer)).Methods("GET")
-	subrouter.HandleFunc("/job_offers/{id}/files", solverServer.jobOfferDownloadFiles).Methods("GET")
+	subrouter.HandleFunc("/job_offers/{id}/files", http.GetHandler(solverServer.jobOfferDownloadFiles)).Methods("GET")
 
 	subrouter.HandleFunc("/resource_offers", http.GetHandler(solverServer.getResourceOffers)).Methods("GET")
 	subrouter.HandleFunc("/resource_offers", http.PostHandler(solverServer.addResourceOffer)).Methods("POST")
@@ -89,7 +89,7 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 	subrouter.HandleFunc("/deals", http.GetHandler(solverServer.getDeals)).Methods("GET")
 	subrouter.HandleFunc("/deals/{id}", http.GetHandler(solverServer.getDeal)).Methods("GET")
 
-	subrouter.HandleFunc("/deals/{id}/files", solverServer.downloadFiles).Methods("GET")
+	subrouter.HandleFunc("/deals/{id}/files", http.GetHandler(solverServer.downloadFiles)).Methods("GET")
 	subrouter.HandleFunc("/deals/{id}/files", solverServer.uploadFiles).Methods("POST")
 
 	subrouter.HandleFunc("/deals/{id}/result", http.GetHandler(solverServer.getResult)).Methods("GET")
@@ -121,7 +121,7 @@ func (solverServer *solverServer) ListenAndServe(ctx context.Context, cm *system
 	anurarouter.HandleFunc("/job_offers", http.PostHandler(solverServer.addJobOffer)).Methods("POST")
 	anurarouter.HandleFunc("/job_offers", http.GetHandler(solverServer.getJobOffers)).Methods("GET")
 	anurarouter.HandleFunc("/job_offers/{id}", http.GetHandler(solverServer.getJobOffer)).Methods("GET")
-	anurarouter.HandleFunc("/job_offers/{id}/files", solverServer.jobOfferDownloadFiles).Methods("GET")
+	anurarouter.HandleFunc("/job_offers/{id}/files", http.GetHandler(solverServer.jobOfferDownloadFiles)).Methods("GET")
 
 	// this will fan out to all connected web socket connections
 	// we read all events coming from inside the solver controller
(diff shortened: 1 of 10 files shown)


pkg/solver/server.go
```diff
@@ -520,50 +524,113 @@ func (solverServer *solverServer) updateTransactionsMediator(payload data.DealTr
 *
 */
 
-func (solverServer *solverServer) downloadFiles(res corehttp.ResponseWriter, req *corehttp.Request) {
+type EmptyResponse struct{}
```
narbs91, line 544: I think it might be cleaner to just return `nil` instead of an empty response, what do you think?  Alternatively, you could make this a `StatusResponse` object that contains an HTTP status object with the message and the statusCode

pkg/solver/server.go
```diff
@@ -520,50 +524,113 @@ func (solverServer *solverServer) updateTransactionsMediator(payload data.DealTr
...
+
+func (solverServer *solverServer) downloadFiles(res corehttp.ResponseWriter, req *corehttp.Request) (EmptyResponse, error) {
```
narbs91, line 546: If we removed the EmptyResponse, then we'd just need to change the method signature to look like this (essentially becoming a void func)

```suggestion
func (solverServer *solverServer) downloadFiles(res corehttp.ResponseWriter, req *corehttp.Request) error {
```

pkg/solver/server.go
```diff
@@ -520,50 +524,113 @@ func (solverServer *solverServer) updateTransactionsMediator(payload data.DealTr
...
+	var targetFile os.DirEntry
+	for _, file := range files {
+		info, err := file.Info()
 		if err != nil {
-			log.Error().Err(err).Msgf("error checking signature")
-			return &http.HTTPError{
-				Message:    errors.New("not authorized").Error(),
-				StatusCode: corehttp.StatusUnauthorized,
-			}
+			continue
```
narbs91, line 611: Is this supposed to be a `continue` here or a `break`?

pkg/solver/server.go
```diff
@@ -520,50 +524,113 @@ func (solverServer *solverServer) updateTransactionsMediator(payload data.DealTr
...
-		signerAddress, err := http.CheckSignature(req)
+	// Check authorization
+	signerAddress, err := http.CheckSignature(req)
+	if err != nil {
+		return EmptyResponse{}, &http.HTTPError{
+			Message:    "not authorized",
+			StatusCode: corehttp.StatusUnauthorized,
+		}
+	}
+	if signerAddress != deal.JobCreator {
```
narbs91, line 573: Let's do a toLower on this check just to avoid funky stuff with formatting of wallet addresses coming

pkg/solver/server.go
```diff
@@ -645,110 +712,71 @@ func (solverServer *solverServer) uploadFiles(res corehttp.ResponseWriter, req *
...
+			Message:    errors.New("not authorized").Error(),
+			StatusCode: corehttp.StatusUnauthorized,
 		}
 	}
 
-	if targetFile == nil {
-		return &http.HTTPError{
-			Message:    "no regular files found in directory",
-			StatusCode: corehttp.StatusNotFound,
+	if signerAddress != jobOffer.JobCreator {
```
narbs91, line 762: Same thing about toLower here

pkg/solver/stats/jobrun.go
```diff
@@ -0,0 +1,67 @@
...
+		ModuleID:                  shortcuts.GetShortcut(deal.Deal.JobOffer.Module.Repo, deal.Deal.JobOffer.Module.Hash),
+		ResourceProvider:          deal.ResourceProvider,
+		JobCreator:                deal.JobCreator,
+		TotalDurationMilliseconds: float64(jobRunTime),
+		ExtraData:                 extraData,
+		JobOffer:                  deal.JobOffer,
+		ResourceOffer:             deal.ResourceOffer,
+	}
+
+	_, err = http.PostRequest[JobRun, JobRun](stat.ClientOptions, "/job-run", jobRun)
```
narbs91, line 48: Does the Stats API return a response at all?  Like is there something that we can use from that response that we might be able to return instead of having this function return nil?

pkg/solver/stats/jobrun.go
```diff
@@ -0,0 +1,67 @@
...
+	JobCreator                string          `json:"jobCreator"`
+	TotalDurationMilliseconds float64         `json:"totalDuration_ms"`
+	ExtraData                 json.RawMessage `json:"extraData"`
+	JobOffer                  string          `json:"jobOffer"`
+	ResourceOffer             string          `json:"resourceOffer"`
+}
+
+// Stats API implementation
+
+func (stat *HTTPStats) PostJobRun(store store.SolverStore, deal *data.DealContainer) error {
```
narbs91: This should probably take in a JobOffer object instead of the SolverStore to separate concerns.  In fact think this method should take in a request object containing whatever the Stats API is expecting in the endpoint

pkg/solver/stats/reputation.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// The reputation builder constructs reputation from
+// reputation events. For now, it does not expose runtime
+// millis or module ID.
+type ReputationBuilder struct {
+	reputation Reputation
+}
+
+func NewReputationBuilder() *ReputationBuilder {
```
narbs91, line 25: Awesome, love me some builders!

pkg/solver/stats/reputation.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	return nil
+}
+
+// Noop implementation
+
+func (stat *NoopStats) PostReputation(address string, reputation Reputation) error {
+	return nil
+}
```
narbs91, line 69: I think this and the endpoint related code in `jobrun.go` should be in the `stats_client.go` file so that we keep all the client related code in one place.  Then we can keep the `jobrun` and `reputation` as the model/builder classes for the objects and perhaps even the request objects for those respective endpoints.  So we basically create a models package under stats to house these and then have `stats_client.go` in the root of the stats package

narbs91 (commented):
Looks good overall, just a few questions and some comments around class/package creation

pkg/solver/stats/jobrun.go
bgins: The stats API echoes back the `JobRun`: https://github.com/Lilypad-Tech/stats-api/blob/6915fcaebf8b5c19988d6769208f7462e124eea4/handlers/job_run.go#L34

The expected response body is defined in `http.PostRequest` wrapper as `ResultType` using generics:

https://github.com/Lilypad-Tech/lilypad/blob/eaa55093790fda1b065c0a9a30081d65e6989f9d/pkg/http/utils.go#L418-L422

We validate the result type in the response:

https://github.com/Lilypad-Tech/lilypad/blob/eaa55093790fda1b065c0a9a30081d65e6989f9d/pkg/http/utils.go#L440
https://github.com/Lilypad-Tech/lilypad/blob/eaa55093790fda1b065c0a9a30081d65e6989f9d/pkg/http/utils.go#L464

We could return the `JobRun` from the response, but if we don't have a use for it, we may end up ignoring it at the `PostJobRun` call site.

narbs91: Gotcha, if there's no use for the response object then we can keep it as

pkg/solver/server.go
bgins: This PR doesn't alter this functionality, but it's worth considering!

We loop over files some job outputs directory (for example `/tmp/lilypad/data/downloaded-files/QmP7FroLkDFgyLSos5nRYVAitbYFL7CQxxo3qqDqeH7z5U`), looking for job outputs to download. For each file, we check `file.Info()`. If `file.Info()` returns an error we continue.

What are the conditions where we might encounter an error? The `file.Info()` documentation says:

> Info returns the FileInfo for the file or subdirectory described by the entry. The returned FileInfo may be from the time of the original directory read or from the time of the call to Info. If the file has been removed or renamed since the directory read, Info may return an error satisfying errors.Is(err, ErrNotExist). If the entry denotes a symbolic link, Info reports the information about the link itself, not the link’s target.

So we get an error if the file went missing or was renamed after we read the directory. Continuing here means we will attempt to return any other files that are still there, essentially returning a partial output. Seem right?

This definitely seems like an edge case, but maybe we should report an error if the edge case comes up?

pkg/solver/server.go
bgins: We may not be able to return `nil` because our `http.PostHandler` needs to infer a type, and not sure if it would be possible to use a pointer type here.

Returning `StatusResponse` reads better, but my understanding is the client won't receive it. The client is expecting a file stream, so I think they will ignore the HTTP body.

bgins: We decided to stay with `EmptyResponse` with better documentation stating its purpose: e60b1c768dd0bf189920e83d841674269af984d8

pkg/solver/server.go
bgins: We decided to stay with `EmptyResponse` because our `http.GetHandler` wrapper requires a type to fill the generic.

pkg/solver/server.go
bgins: We think for now continuing best effort is our best option at this point, but we've added a log so we are at least record when this happens: f06a91c185dd2e16d24c9f15467b91d71712cdd4

pkg/solver/server.go
bgins: We've decided to log when this happens while we consider a general approach to this issue: 4fbd8f24c75865b7213b4b637ba091159c19f0a6

pkg/solver/server.go
bgins: We've decided to log when this happens while we consider a general approach to this issue: 4fbd8f24c75865b7213b4b637ba091159c19f0a6

pkg/solver/stats/reputation.go
bgins: We'll keep the current structure for now, but we may break it up if the stats implementation grows in the future.

pkg/solver/stats/jobrun.go
bgins: After a great design conversation, we decided to pass in the job offer and avoid passing in the store. While implementing, I realized we had the job offer on the `Deal` struct so we can skip querying for the job offer entirely!

Change in this commit: 3cc45ffa34fb3cd1e440ba6f1b842bcf1257dfbe

narbs91 (approved):
LGTM!  Great work @bgins 🙌

PR #526, opened 2025-03-03
feat: Add lilynext flag

bgins:
### Summary

- [x] Add a global `lilynext` flag
- [x] Extract the `lilynext` flag in the solver, resource provider, run, and job creator commands

This pull request adds a `lilynext` feature flag to the gate the new protocol while we implement and stabilize it. The flag is available globally and consumed by commands that may implement the new protocol.

### Test plan

Run the solver with the `lilynext` flag:

```
./stack solver --lilynext
```

The solver should log a line at startup like:

```
2025-03-03T11:09:18-08:00 INF cmd/lilypad/solver.go:35 > 🍃 Running the new lilypad protocol
```

Running the solver without the `lilynext` flag should not log this line.

The same should hold true for the resource provider, run command, and job creator proxy.

### Details

We will need make the feature flag available throughout our code, wherever we implement new protocol.

Although we do not actively support the onchain job creator, we pass the flag to it because it shares functionality with the run command, so it may need it.

### Related issues or PRs

Epic: https://github.com/Lilypad-Tech/internal/issues/407

### DIFF
--- a/cmd/lilypad/jobcreator.go
+++ b/cmd/lilypad/jobcreator.go
@@ -20,11 +20,12 @@ func newJobCreatorCmd() *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 
 			network, _ := cmd.Flags().GetString("network")
+			lilynext, _ := cmd.Flags().GetBool("lilynext")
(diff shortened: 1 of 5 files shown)


narbs91 (approved):
LGTM!

PR #528, opened 2025-03-05
feat: Isolate job output in cli

hollygrimm:
### Summary

- [X] Add JobOfferSubscription struct to support optional job ID filtering
- [X] Modify SubscribeToJobOfferUpdates to return an unsubscribe function
- [X] Introduce new SubscribeToJobOfferUpdatesWithFilter method
- [X] Update job offer update mechanism to respect job ID filters
- [X] Refactor JobCreator to use an interface for better extensibility

By isolating job output, tools like the Discord Observer can run jobs in parallel and get the appropriate job output.

### Task/Issue reference

Closes: [404](https://github.com/Lilypad-Tech/internal/issues/404)

### Test plan

1. Start the stack. Run the integration tests: `./stack integration-tests`

2. Run a job and compare the logs in the cli with the logs in this pull request.


pkg/jobcreator/run.go
```diff
@@ -60,6 +58,26 @@ func RunJob(
 	ctx.Ctx = c
 	defer span.End()
 
+	span.AddEvent("add_job_offer.start")
```
bgins: We can remove this span event and use this one: https://github.com/Lilypad-Tech/lilypad/blob/dcf62f353cd764a4f1a212c81e9e94dea12012d7/pkg/jobcreator/run.go#L81

Lexically, it looks like we should put it before the subscription span events, but testing it out, the subscription events won't be posted until after the job offer is posted.

pkg/jobcreator/controller.go
```diff
@@ -91,8 +102,26 @@ func (controller *JobCreatorController) AddJobOffer(offer data.JobOffer) (data.J
...
+func (controller *JobCreatorController) SubscribeToJobOfferUpdates(sub JobOfferSubscriber) func() {
+	return controller.SubscribeToJobOfferUpdatesWithFilter(sub, "")
+}
+
+// SubscribeToJobOfferUpdatesWithFilter subscribes to job offer updates with an optional job ID filter
+// If jobID is not empty, only updates for that job will be sent to the subscriber
+// Returns a function that can be called to unsubscribe
+func (controller *JobCreatorController) SubscribeToJobOfferUpdatesWithFilter(sub JobOfferSubscriber, jobID string) func() {
+	id := controller.nextSubID
+	controller.nextSubID++
```
bgins: What would you think about making `controller.jobOfferSubscriptions` a `map[string]JobOfferSubscription` where the key is the job offer ID? 

The `JobOfferSubscription` struct also has it as a field, so it feels a little repetitive. But it looks like what we want is uniqueness, and it would let us avoid the counter.

pkg/jobcreator/controller.go
```diff
@@ -138,8 +167,13 @@ func (controller *JobCreatorController) subscribeToSolver() error {
...
 			}
 			metricsDashboard.TrackJobOfferUpdate(*ev.JobOffer)
+
+			// Make a copy of the subscriptions to avoid modification during iteration
 			for _, sub := range controller.jobOfferSubscriptions {
-				go sub(*ev.JobOffer)
+				// If JobID filter is set, only send updates for that job
+				if sub.JobID == "" || sub.JobID == ev.JobOffer.JobOffer.ID {
+					sub.Callback(*ev.JobOffer)
+				}
```
bgins: Yeah awesome! 🎉

pkg/jobcreator/jobcreator.go
```diff
@@ -43,46 +43,56 @@ type JobCreatorOptions struct {
...
-	web3SDK    *web3.Web3SDK
-	options    JobCreatorOptions
+type JobCreator interface {
+	Start(ctx context.Context, cm *system.CleanupManager) chan error
+	GetJobOfferFromOptions(options JobCreatorOfferOptions) (data.JobOffer, error)
+	AddJobOffer(offer data.JobOffer) (data.JobOfferContainer, error)
+	SubscribeToJobOfferUpdates(sub JobOfferSubscriber) func()
+	SubscribeToJobOfferUpdatesWithFilter(sub JobOfferSubscriber, jobID string) func()
+	GetResult(dealId string) (data.Result, error)
+}
```
bgins, line 53: Nice one, cleans things up and sets us up for future implementers. 💯

bgins (approved):
Looks good! Left a couple comments with some suggested changes.

pkg/jobcreator/controller.go
hollygrimm: great catch! this has been done

pkg/jobcreator/run.go
hollygrimm: Yes, this artifact was left after I moved the subscription code to before the job offer. Now removed

hollygrimm:
@bgins yes, since there is only ever one subscription per job id, it makes sense that we remove the nextSubID counter. thanks for your review

PR #539, opened 2025-03-14
feat: Module Loading Improvements

kelindi:
For new clones:

- We only acquire the lock when we need to clone a new repository
- Added a double-check after acquiring the lock to see if another process created the repo while we were waiting
- Properly release the lock in all error cases
- The lock is held only during the clone/checkout process

For existing repositories:

- Don't acquire a lock
- We simply verify that the HEAD is at the correct commit hash
- If the repository exists but is corrupt or has the wrong hash, we remove it and retry
- This makes the common case (using an already cloned repo) very fast and lock-free

Testing:
```bash
go test -v -tags=unit ./pkg/module -run TestTryLock
```


bgins:
Could we remove this function? Looks like we aren't using it any more: https://github.com/Lilypad-Tech/lilypad/blob/58b7d955476f3f6db6ca6945cae1ab8db5860ff6/pkg/module/utils.go#L61-L73

pkg/module/utils.go
```diff
@@ -68,72 +90,412 @@ func ProcessModule(module data.ModuleConfig) (data.ModuleConfig, error) {
...
-		// this means there is no hash in the repo
-		// so let's clean it up and clone it again
-		err = os.RemoveAll(repoDir)
+}
+
+func CloneModule(module data.ModuleConfig) (*git.Repository, error) {
+	maxRetries := 3
+	var lastErr error
+
+	for retry := 0; retry < maxRetries; retry++ {
```
bgins: Could we convert this to a `range` loop over `maxRetries`?

pkg/module/utils.go
```diff
@@ -68,72 +90,412 @@ func ProcessModule(module data.ModuleConfig) (data.ModuleConfig, error) {
...
+						Str("path", lockFile).
+						Str("content", string(content)).
+						Err(scanErr).
+						Msg("Could not parse lock file, removing stale lock")
+					os.Remove(lockFile)
+					continue
+				}
+
+				// Check if the lock is too old (> 10 minutes)
+				lockTime, parseErr := time.Parse(time.RFC3339, timestamp)
```
bgins, line 244: Should we handle the case where a `parseErr` is returned?

pkg/module/utils.go
```diff
@@ -249,11 +622,131 @@ func LoadModule(module data.ModuleConfig, inputs map[string]string) (*data.Modul
...
+		Str("repo", repoURL).
+		Str("reference", reference).
+		Str("tmpDir", tmpDir).
+		Msg("Resolving git reference to commit hash")
+
+	// Clone with retries
+	var repo *git.Repository
+	var cloneErr error
+
+	for retry := 0; retry < 3; retry++ {
```
bgins: Same as above. Could we modernize to a range loop?

pkg/module/utils_test.go
```diff
@@ -57,3 +61,154 @@ func TestSubst(t *testing.T) {
...
+		// Clean up
+		unlock()
+	})
+
+	// Test 4: Can acquire lock after old lock is removed (timeout based)
+	t.Run("Old Lock", func(t *testing.T) {
+		lockFile := filepath.Join(tempDir, ".lilypad.lock")
+
+		// Create an old lock file with valid PID but old timestamp
+		oldTime := time.Now().Add(-11 * time.Minute).Format(time.RFC3339)
```
bgins, line 135: Ha great trick adding -11, love it. ❤️

pkg/module/utils_test.go
```diff
@@ -57,3 +61,154 @@ func TestSubst(t *testing.T) {
...
+		concurrentDir := filepath.Join(t.TempDir(), "concurrent")
+		err := os.MkdirAll(concurrentDir, 0755)
+		assert.NoError(t, err, "Should create concurrent test directory")
+
+		// This will track which goroutine got the lock
+		gotLock := make(chan int, 5)
+		var wg sync.WaitGroup
+
+		// Launch 5 goroutines to try to get the lock
+		for i := 0; i < 5; i++ {
```
bgins: Could we convert this one for a for range loop as well?

pkg/module/utils_test.go
```diff
@@ -57,3 +61,154 @@ func TestSubst(t *testing.T) {
...
+
+	// Test 5: Concurrent locking attempts should work correctly
+	t.Run("Concurrent Locking", func(t *testing.T) {
+		// Create multiple directories to avoid test interference
+		concurrentDir := filepath.Join(t.TempDir(), "concurrent")
+		err := os.MkdirAll(concurrentDir, 0755)
+		assert.NoError(t, err, "Should create concurrent test directory")
+
+		// This will track which goroutine got the lock
+		gotLock := make(chan int, 5)
```
bgins: Might be good to put this `5` in a variable if we expect it to be the same used in the loop a couple lines below.

hollygrimm:
@bgins Is this overkill?
```go
// ModuleTimeoutConfig holds all timeout configuration values
type ModuleTimeoutConfig struct {
	// Lock related timeouts
	DefaultLockTimeout  time.Duration // Default timeout for acquiring lock
	ExtendedLockTimeout time.Duration // Extended timeout for lock acquisition during clone
	LockTickInterval    time.Duration // Interval between lock acquisition attempts
	StaleLockThreshold  time.Duration // Time after which a lock is considered stale

	// Git operation related timeouts and retries
	MaxRetries             int           // Maximum number of retries for git operations
	RetryDelayMultiplier   time.Duration // Base delay multiplier for retries
	NetworkRetryMultiplier time.Duration // Additional delay for network-related errors
}

// DefaultTimeoutConfig provides default values for all timeouts
var DefaultTimeoutConfig = ModuleTimeoutConfig{
	DefaultLockTimeout:     30 * time.Second,
	ExtendedLockTimeout:    60 * time.Second,
	LockTickInterval:       100 * time.Millisecond,
	StaleLockThreshold:     10 * time.Minute,
	MaxRetries:             3,
	RetryDelayMultiplier:   2 * time.Second,
	NetworkRetryMultiplier: 5 * time.Second,
}

// Current timeout configuration, can be modified at runtime
var TimeoutConfig = DefaultTimeoutConfig

// ConfigureTimeouts allows users to modify the timeout configuration
// Partial configurations are supported - only the provided fields will be updated
func ConfigureTimeouts(config ModuleTimeoutConfig) {
	// Only update non-zero values
	if config.DefaultLockTimeout > 0 {
		TimeoutConfig.DefaultLockTimeout = config.DefaultLockTimeout
	}

	if config.ExtendedLockTimeout > 0 {
		TimeoutConfig.ExtendedLockTimeout = config.ExtendedLockTimeout
	}

	if config.LockTickInterval > 0 {
		TimeoutConfig.LockTickInterval = config.LockTickInterval
	}

	if config.StaleLockThreshold > 0 {
		TimeoutConfig.StaleLockThreshold = config.StaleLockThreshold
	}

	if config.MaxRetries > 0 {
		TimeoutConfig.MaxRetries = config.MaxRetries
	}

	if config.RetryDelayMultiplier > 0 {
		TimeoutConfig.RetryDelayMultiplier = config.RetryDelayMultiplier
	}

	if config.NetworkRetryMultiplier > 0 {
		TimeoutConfig.NetworkRetryMultiplier = config.NetworkRetryMultiplier
	}

	log.Debug().
		Interface("timeoutConfig", TimeoutConfig).
		Msg("Updated module timeout configuration")
}

// ResetTimeouts resets all timeouts to default values
func ResetTimeouts() {
	TimeoutConfig = DefaultTimeoutConfig
	log.Debug().Msg("Reset module timeout configuration to defaults")
}
```
override with this call:
```go
module.ConfigureTimeouts(module.ModuleTimeoutConfig{
    ExtendedLockTimeout: 2 * time.Minute,
    MaxRetries: 5,
})
```

bgins (approved):
Looks good! Let's send it 🎉

PR #544, opened 2025-03-18
feat: Add minimum version check

### LINKED ISSUE #543: Add minimum version check
## General Description

Add a minimum version check to the solver so we can ensure that job creators and resource providers stay up to date.

This functionality affects solver, resource provider, and job creator.

The solver should implement a minimum version check. The minimum version should be configurable and fall back to the major component of the solver version.

bgins:
### Summary

- [x] Add `SERVER_MINIMUM_VERSION` option and version config struct
- [x] Add minimum version check
- [x] Add version config and version check tests
- [x] Check version on resource offer and job offer post 

This pull request adds a minimum version check so we can ensure that job creators and resource providers stay up to date at a `SERVER_MINIMUM_VERSION` or newer.

### Task/Issue reference

Closes: #543 

### Test plan

We include the version at compile time, so testing this pull request will require a few compiled binaries. Start by downloading the `v2.13.0`, `v2.14.0`, and `v2.15.0` binaries from our releases page: https://github.com/Lilypad-Tech/lilypad/releases. Rename the binaries to `lilypad-v2.13.0`, `lilypad-v2.14.0`, and `lilypad-v2.15.0`

Start the base services. Start the solver with a `v2.14.0` minimum version.

```sh
./stack solver --server-minimum-version v2.14.0
```

We run the solver from source for this portion of testing because we are testing the `SERVER_MINIMUM_VERSION` setting and the compiled version does not matter.

> [!CAUTION]
> In this next portion of testing we use development-only private keys. Do not use these keys for production or any use case of consequence. Future reader you have been warned. 🔥 

Start a `v2.15.0` resource provider using our local dev private key.

```sh
WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a ./lilypad-v2.15.0 resource-provider --network dev --disable-pow true
```

Check the solver and the resource offer should increment to one resource offer.

Stop the resource provider and start the `v2.14.0` resource provider.

```sh
WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a ./lilypad-v2.14.0 resource-provider --network dev --disable-pow true
```

Once again, the resource offer should be accepted by the solver.

Stop the resource provider and start the `v2.13.0` resource provider.

```sh
WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a ./lilypad-v2.13.0 resource-provider --network dev --disable-pow true
```

The solver should start logging unsupported version errors and the resource provider will crash with the same error message after retries.

#### Semantic version

We fall back to a major version derived from the solver version when the minimum version is not set. To test this, we need to compile a binary for the solver at `v2.16.0` and another binary before `v2.0.0` for the resource provider.

```sh
go build -v -o lilypad-v2.16.0 -ldflags="-X 'github.com/lilypad-tech/lilypad/pkg/system.Version=v2.16.0'"
go build -v -o lilypad-v1.52.0 -ldflags="-X 'github.com/lilypad-tech/lilypad/pkg/system.Version=v1.52.0'"
```

Set the binaries to executable and then start the solver without a minimum version. The easiest way to do this is to temporarily update our `stack` solver command. Change this line:

https://github.com/Lilypad-Tech/lilypad/blob/922c3e79ced16022042affbca221b5c52389dfef/stack#L211

to

```sh
./lilypad-v2.16.0 solver --network dev "$@"
```

Then start the solver with `./stack solver`. This preserves our full local development configuration.

Now start `v1.52.0` resource provider:

```sh
WEB3_PRIVATE_KEY=0x47e179ec197488593b187f80a00eb0da91f1b9d0b13f8733639f19c30a34926a ./lilypad-v1.52.0 resource-provider --network dev --disable-pow true
```

The resource offer should be rejected because it is before the `v2` major version.

### Details

All versions assume a `v` prefix and will fail validation if it is missing.

#### Configs

The mininmum version can be set with an environment variable or a CLI option:

```sh
SERVER_MINIMUM_VERSION=v2.14.0 ./stack solver
./stack solver --server-minimum-version v2.14.0
```

#### Semantic versioning

The minimum version is a convenience until we move to strict semantic versioning. Once we move to strict semantic versioning, we can drop the `SERVER_MINIMUM_VERSION` config and use the solver version to determine the appropriate supported major version as minimum.

#### Local development

The solver does not have a version in local development. The `SERVER_MINIMUM_VERSION` config can be set to enforce a minimum, but the common case should avoid this because job creators and resource providers will also not have a version.

### DIFF
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -351,6 +351,11 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
 	versionHeader, _ := http.GetVersionFromHeaders(req)
 	minVersion, ok := server.versionConfig.IsSupported(versionHeader)
 	if !ok {
+		server.log.Debug().Str("cid", jobOffer.ID).
+			Str("address", jobOffer.JobCreator).
+			Str("version", versionHeader).
+			Str("minVersion", minVersion).
+			Msg("job offer rejected because job creator is running an unsupported version")
 		return nil, fmt.Errorf("Please update to minimum supported version %s or newer: https://github.com/Lilypad-Tech/lilypad/releases", minVersion)
 	}
 
@@ -397,6 +402,11 @@ func (server *solverServer) addResourceOffer(resourceOffer data.ResourceOffer, r
 	versionHeader, _ := http.GetVersionFromHeaders(req)
 	minVersion, ok := server.versionConfig.IsSupported(versionHeader)
 	if !ok {
+		server.log.Debug().Str("cid", resourceOffer.ID).
+			Str("address", resourceOffer.ResourceProvider).
+			Str("version", versionHeader).
+			Str("minVersion", minVersion).
+			Msg("resource offer rejected because resource provider is running an unsupported version")
 		return nil, fmt.Errorf("Please update to minimum supported version %s or newer: https://github.com/Lilypad-Tech/lilypad/releases", minVersion)
 	}
 
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -30,12 +30,13 @@ import (
 )
 
 type solverServer struct {
-	options    http.ServerOptions
-	controller *SolverController
-	store      store.SolverStore
-	stats      stats.Stats
-	services   data.ServiceConfig
-	log        *zerolog.Logger
+	options       http.ServerOptions
+	controller    *SolverController
+	store         store.SolverStore
+	stats         stats.Stats
+	services      data.ServiceConfig
+	versionConfig *system.VersionConfig
+	log           *zerolog.Logger
 }
 
 func NewSolverServer(
@@ -43,14 +44,16 @@ func NewSolverServer(
 	controller *SolverController,
 	store store.SolverStore,
 	stats stats.Stats,
+	versionConfig *system.VersionConfig,
 	services data.ServiceConfig,
 ) (*solverServer, error) {
 	server := &solverServer{
-		options:    options,
-		controller: controller,
-		store:      store,
-		stats:      stats,
-		log:        system.GetLogger(system.SolverService),
+		options:       options,
+		controller:    controller,
+		store:         store,
+		stats:         stats,
+		versionConfig: versionConfig,
+		log:           system.GetLogger(system.SolverService),
 	}
 
 	metricsDashboard.Init(services.APIHost)
--- a/pkg/solver/server.go
+++ b/pkg/solver/server.go
@@ -348,6 +348,12 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
 		return nil, fmt.Errorf("job creator address does not match signer address")
 	}
 
+	versionHeader, _ := http.GetVersionFromHeaders(req)
+	minVersion, ok := server.versionConfig.IsSupported(versionHeader)
+	if !ok {
+		return nil, fmt.Errorf("Please update to minimum supported version %s or newer: https://github.com/Lilypad-Tech/lilypad/releases", minVersion)
+	}
+
 	offerRecent := isTimestampRecent(jobOffer.CreatedAt, server.options.AccessControl.OfferTimestampDiffSeconds*1000)
 	if !offerRecent {
 		server.log.Debug().Str("cid", jobOffer.ID).Str("address", jobOffer.JobCreator).Msg("job offer rejected because timestamp was not recent")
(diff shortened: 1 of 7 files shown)


logan-wrld:
Hey @bgins! Awesome testing instructions! The preparation and the first two steps(startup a v2.14 Solver using the minimum version cmd and a v2.15 RP And then... a 2.14 RP) were completed without any issues. 

After that, I was able to activate a v2.13 RP connected to the same solver(`./stack solver --server-minimum-version v2.14.0`) **without** the RP crashing. 

I left the RP running with the solver for over 30 minutes to see if the RP eventually timed out(it did not).

- From your instructions, the RP should have crashed correct? 

Below is a screenshot of the RP connecting to the Solver: 
Screenshot from 2025-03-19 01-58-44

Just to see what happens when running a cowsay in this configuration, the Solver gives me this error `...server.go:2220 > error="Please update to minimum supported version v2.14.0 or newer: https://github.com/Lilypad-Tech/lilypad/releases" method POST=/api/v1/job_offers` 

Screenshot from 2025-03-19 02-02-23

bgins:
> The preparation and the first two steps(startup a v2.14 Solver using the minimum version cmd and a v2.15 RP And then... a 2.14 RP) were completed without any issues.
> 
> After that, I was able to activate a v2.13 RP connected to the same solver(`./stack solver --server-minimum-version v2.14.0`) **without** the RP crashing.

The solver command in the second step looks correct, and the resource provider should have crashed well before 30 minutes. The solver should also start logging not supported messages immediately.

> Just to see what happens when running a cowsay in this configuration, the Solver gives me this error `...server.go:2220 > error="Please update to minimum supported version v2.14.0 or newer: https://github.com/Lilypad-Tech/lilypad/releases" method POST=/api/v1/job_offers`

Yeah, this is what the unsupported version case should look like.

Let's talk through the testing later today. Would like to check a couple details in your setup, and it'll be faster to discuss it live.

pkg/solver/server.go
```diff
@@ -345,6 +348,12 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
 		return nil, fmt.Errorf("job creator address does not match signer address")
 	}
 
+	versionHeader, _ := http.GetVersionFromHeaders(req)
+	minVersion, ok := server.versionConfig.IsSupported(versionHeader)
+	if !ok {
+		return nil, fmt.Errorf("Please update to minimum supported version %s or newer: https://github.com/Lilypad-Tech/lilypad/releases", minVersion)
```
narbs91, line 359: Let's add a log statement here too

pkg/solver/server.go
```diff
@@ -388,6 +394,12 @@ func (server *solverServer) addResourceOffer(resourceOffer data.ResourceOffer, r
 		}
 	}
 
+	versionHeader, _ := http.GetVersionFromHeaders(req)
+	minVersion, ok := server.versionConfig.IsSupported(versionHeader)
+	if !ok {
+		return nil, fmt.Errorf("Please update to minimum supported version %s or newer: https://github.com/Lilypad-Tech/lilypad/releases", minVersion)
```
narbs91, line 410: Same comment here around adding a log statement

pkg/system/version.go
```diff
@@ -1,4 +1,61 @@
...
+	}
+
+	if minimumVersion != "" && !semver.IsValid(minimumVersion) {
+		return nil, fmt.Errorf("invalid minimum version: %s", minimumVersion)
+	}
+
+	if Version != "" &&
+		minimumVersion != "" &&
+		semver.Compare(minimumVersion, Version) > 0 {
+		return nil, fmt.Errorf("minimum version %s is greater than embedded version %s", minimumVersion, Version)
```
narbs91: Let's add log statements inside these blocks at least initially to help us debug/troubleshoot

pkg/solver/server.go
bgins: Yeah good call! Added in db9caa2b0b0609898642732b33941014cba8b448

pkg/solver/server.go
bgins: Added in db9caa2b0b0609898642732b33941014cba8b448

pkg/system/version.go
bgins: Same, added in db9caa2b0b0609898642732b33941014cba8b448

narbs91 (approved):
LGTM!

logan-wrld (approved):
AMAZING!!! TYSMMM

PR #545, opened 2025-03-20
fix: Prevent duplicate job offers

bgins:
### Summary

- [x] Add job offer nonce

This pull request adds a nonce to job offers to ensure they are unique.

### Test plan

Start the stack. Run multiple jobs in parallel using the same input. Job offers with matching timestamps should remain have unique CIDs with the new nonce.

### Details

Job offers include a `CreatedAt` field that acts as a nonce to distinguish them, but we have found that job offers created in parallel with the same input may not be unique when timestamps collide. We have added a nonce to ensure job offers are not duplicated.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -190,6 +190,7 @@ require (
 	github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0 // indirect
 	github.com/magiconair/properties v1.8.7 // indirect
 	github.com/marten-seemann/tcp v0.0.0-20210406111302-dfbc87cc63fd // indirect
+	github.com/matoous/go-nanoid/v2 v2.1.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.15 // indirect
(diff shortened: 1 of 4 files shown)


pkg/solver/controller.go
hollygrimm, line 1: Could a race condition still occur with this code?
- Two processes calculate the same ID (using data.GetJobOfferID)
- Both processes check if the offer exists (both see it doesn't exist yet)
- Both processes try to add it to the store

Instead of this code, can we put a unique contraint on this field in the database?

hollygrimm: Alternately, we can fix the JC CalculateCID by adding a nanoid. This would guarantee uniqueness. Would this mess anything up downstream?
```
func CalculateCID(v interface{}) (string, error) {
    // Generate a nanoid
    nanoID, err := nanoid.New()
    if err != nil {
        return "", err
    }
    
    // Create a wrapper that includes both the original data and the nanoid
    wrapper := struct {
        Data   interface{} `json:"data"`
        NanoID string      `json:"nano_id"`
    }{
        Data:   v,
        NanoID: nanoID,
    }
    
    // Serialize the wrapper to JSON
    data, err := json.Marshal(wrapper)
    if err != nil {
        return "", err
    }
    
    // Create a Dag Node from the JSON bytes
    node := mdag.NodeWithData(data)
    
    // Compute CID of the Dag Node
    c := node.Cid()
    
    return c.String(), nil
}
```

bgins: > Could a race condition still occur with this code?
> 
> * Two processes calculate the same ID (using data.GetJobOfferID)
> * Both processes check if the offer exists (both see it doesn't exist yet)
> * Both processes try to add it to the store
> 
> Instead of this code, can we put a unique contraint on this field in the database?

Yes, good point. This could happen. A database unique constraint could work.

> Alternately, we can fix the JC CalculateCID by adding a nanoid. This would guarantee uniqueness. Would this mess anything up downstream?

We may not want to do this ad hoc when computing the CID. A job creator could calculate their job offer CID for display, and then when it is recalculated on the solver and ends up with a different CID. But maybe adding the nanoid as a nonce on the job offer struct?

hollygrimm (approved):
Looks great, Brian!

PR #547, opened 2025-03-21
feat: setup foundation for new protocol

narbs91:
### Summary

- [x] Introduces the web3 service logger
- [x] Imports the Go Bindings for the new smart contracts
- [x] Adds the appropriate configurations for protocol v2 separating them from the existing protocol integration
- [x] Adds the api_v2.go module to house the new api endpoints according to the new protocol
- [x] Adds the sdk_v2.go module to represent the new sdk for the new protocol

Explain the motivation for making these changes. Does this pull request implement a feature or fix a bug? Is it a docs change or a typo fix?

This PR is the first of many to introduce the new protocol into the lilypad main repo

### Task/Issue reference

Closes: add_link_here

### Test plan

Please describe how reviewers can test the changes in this pull request. If the test plan is challenging to explain in text alone, a short video demonstration may be included.

There should be no change to the existing functionality of Lilypad at this point. To sanity test the changes, follow what is in the [Local Development guide](https://github.com/Lilypad-Tech/lilypad/blob/main/LOCAL_DEVELOPMENT.md) to ensure that the all the necessary services boot up and you are able to run a cowsay job

### Details (optional)

Add any additional details that will help to review this pull request.

Add any related issues or PRs.


pkg/options/web3.go
bgins, line 1: Comment only, no change requested. Noticed this leaves out CLI options for the new contract addresses. This seems like the right call, at least at first when their use will mostly be internal. Possibly we want to leave them out long term because we automatically configure in the embedded configs.

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+	defer span.End()
+
+	rpcs := strings.Split(options.RpcURL, ",")
+	var err error
+	var parsedURL *url.URL
+	var client *ethclient.Client
+	for _, u := range rpcs {
+		parsedURL, err = url.Parse(u)
+		if err != nil {
+			logger.Warn().Msgf("Unable to parse web3 RPC URL: %v", err)
```
bgins:
```suggestion
			logger.Warn().Err(err).Msg("Unable to parse web3 RPC URL")
```
Could we log the error with an `Err` call to add it as a key-value pair?

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+			span.RecordError(errors.New("Unable to parse web3 RPC URL"))
+			continue
+		}
+
+		span.AddEvent("ethclient.dial", trace.WithAttributes(attribute.String("web3.rpc_url", parsedURL.Host)))
+		client, err = ethclient.Dial(u)
+		if err != nil {
+			logger.Warn().Msgf("Failed to connect to %s: %v", parsedURL.Host, err)
```
bgins: Same with this error, and maybe we can use `Str` for the parsed URL host.

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+			span.RecordError(fmt.Errorf("Failed to connect to %s", parsedURL.Host))
+			continue
+		} else {
+			logger.Info().Msgf("Connected to %s", parsedURL.Host)
```
bgins: Same, could use `Str()` here.

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	return client, nil
+}
+
+func (sdk *Web3SDKV2) getBlockNumber() (uint64, error) {
+	var blockNumberHex string
+	err := sdk.Client.Client().Call(&blockNumberHex, "eth_blockNumber")
+	if err != nil {
+		logger.Error().Msgf("error for getBlockNumber: %s", err.Error())
```
bgins: Same, could use `Err` for the error here.

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+}
+
+func (sdk *Web3SDKV2) GetBalance(address string) (*big.Int, error) {
+	// Convert the string address to common.Address
+	ethAddress := common.HexToAddress(address)
+
+	// Get the balance using the converted address
+	balance, err := sdk.Client.BalanceAt(context.Background(), ethAddress, nil)
+	if err != nil {
+		logger.Error().Msgf("error for GetBalance: %s", err.Error())
```
bgins: Same, use `Err`.

pkg/web3/sdk.go
```diff
@@ -285,7 +290,7 @@ func (sdk *Web3SDK) getBlockNumber() (uint64, error) {
 	var blockNumberHex string
 	err := sdk.Client.Client().Call(&blockNumberHex, "eth_blockNumber")
 	if err != nil {
-		log.Error().Msgf("error for getBlockNumber: %s", err.Error())
+		logger.Error().Msgf("error for getBlockNumber: %s", err.Error())
```
bgins:
```suggestion
		sdkLogger.Error().Msgf("error for getBlockNumber: %s", err.Error())
```

pkg/web3/sdk_v2.go
```diff
@@ -0,0 +1,225 @@
...
+		LilypadToken:            lilypadToken,
+		LilypadUser:             lilypadUser,
+		LilypadTokenomics:       lilypadTokenomics,
+	}, nil
+}
+
+func NewContractSDKV2(ctx context.Context, options Web3Options, tracer trace.Tracer) (*Web3SDKV2, error) {
+	displayOpts := options
+	displayOpts.PrivateKey = "*********"
+	logger.Debug().Msgf("NewContractSDK: %+v", displayOpts)
```
bgins:
```suggestion
	logger.Debug().Any("options", displayOpts).Msg("Creating new contract SDK")
```
Zerolog has an `Any` that we may be able to use on the options.

bgins (approved):
Looks good! Left a few suggestions on the logs. Probably not worth making these kinds of changes for the V1 logs.

pkg/web3/sdk_v2.go
narbs91: Yup, indeed seems to work:

image

pkg/options/web3.go
narbs91: Yeah, I was thinking about the CLI options for the contracts but didn't think it made sense and personally would prefer to just go through the config files in the event we need to do so.

narbs91:
> Looks good! Left a few suggestions on the logs. Probably not worth making these kinds of changes for the V1 logs.

Much appreciated! Made the revisions you requested if you want to take one more quick glance

bgins:
> > Looks good! Left a few suggestions on the logs. Probably not worth making these kinds of changes for the V1 logs.
> 
> Much appreciated! Made the revisions you requested if you want to take one more quick glance

Looks good! Let's send it 🙌

PR #554, opened 2025-03-26
fix: Exempt Anura from version check

bgins:
### Summary

- [x] Exempt Anura from version check

We recently added a minimum version check (#544) for job offers and resource offers. This check works for resource providers and job creators versioned in this repo, but it does not work for Anura.

This pull request adds an exemption from the version check for Anura.

### Task/Issue reference

Implements:  #543

### Test plan

Start the base services and solver with:

```sh
SERVER_MINIMUM_VERSION=v2.14.0 ./stack solver
```

This assumes the local dev `SERVER_ANURA_ADDRESSES` set in `.local.dev`: https://github.com/Lilypad-Tech/lilypad/blob/37e86a39122a54260d72d853e7282f1925c0a70c/.local.dev#L33

Remove the `X-Lilypad-Version` header changed recently in Anura (https://github.com/Lilypad-Tech/anura/pull/20/files). The missing version would cause a CLI job offer to be rejected, but Anura should be exempt.

Start Anura and send some jobs. They should post.

### DIFF
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -225,6 +225,30 @@ func CheckAnuraSignature(req *http.Request, approvedAddresses []string) (string,
 	}
 }
 
+func IsAnura(req *http.Request, approvedAddresses []string) bool {
+	serverHeader := req.Header.Get(X_ANURA_SERVER_HEADER)
+	if serverHeader == "" {
+		return false
+	}
+	anuraSignature := req.Header.Get(X_ANURA_SIGNATURE_HEADER)
+	if anuraSignature == "" {
+		return false
+	}
+
+	signatureAddress, err := decodeUserAddress(serverHeader, anuraSignature)
+	if err != nil {
+		return false
+	}
+
+	for _, addr := range approvedAddresses {
+		if strings.EqualFold(signatureAddress, addr) {
+			return true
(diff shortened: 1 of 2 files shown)


pkg/http/utils.go
```diff
@@ -225,6 +225,30 @@ func CheckAnuraSignature(req *http.Request, approvedAddresses []string) (string,
 	}
 }
 
+func IsAnura(req *http.Request, approvedAddresses []string) bool {
+	serverHeader := req.Header.Get(X_ANURA_SERVER_HEADER)
+	if serverHeader == "" {
+		return false
+	}
+	anuraSignature := req.Header.Get(X_ANURA_SIGNATURE_HEADER)
```
kelindi, line 233: I wonder if we should check the signature to make sure its actually from Anura. otherwise people can just add these headers to any request from an older client

bgins: We check if the signature address is in the configured addresses a bit further on: https://github.com/Lilypad-Tech/lilypad/blob/2bb2bd6d557d1b6432d8918492a75064fd0e5b45/pkg/http/utils.go#L243-L247

kelindi (approved):
lgtm

PR #557, opened 2025-03-27
feat: Compute and report job run timings

### LINKED ISSUE #556: Add job lifecycle timings
## General Description

We recently added coarse-grained job run timing in #517. We would also like to record match, execution, and job output retrieval times.

This pull request affects the solver and stats API.

We should record deal creation, job output upload, and job output download times as markers to compute the constituent time spans in the job lifecycle.

bgins:
### Summary

- [x] Add `CreatedAt`, `UploadAt`, and `DownloadAt` fields to deal container
  - [x] Set `CreatedAt` time at deal container creation
  - [x] Set `UploadAt` when resource provider completes job output upload
    - [x] Add  `UpdateDealUploadTime` solver store method
  - [x] Set `DownloadAt` when the job creator completes job output download
    - [x] Add `UpdateDealDownloadTime` solver store method
- [x] Compute and report job run timings to stats API
  - [x] Add job timing and job state fields to solver stats `JobRun`
  - [x] Report timings for successful job runs
  - [x] Report timings for timed out job runs 

We would like to record match, execution, and job output retrieval times to the stats API. The execution times will be used as a baseline for module-specific execution times.

### Task/Issue reference

Closes: #556 

### Test plan

This pull request should be tested alongside this stats API PR https://github.com/Lilypad-Tech/stats-api/pull/5. Start the stats API as described there.

Start the base services and the solver with:

```sh
```

Start a resource provider and run a job. Check the stats API database for the job run timings.

Run two more jobs with interruptions:

- Start a job and stop the resource provider while it is executing the job
- Start a job and stop the job creator after the job has been matched

These test cases should demonstrate timed out job run reports. 

To speed up the timeouts, adjust `JOB_TIMEOUT_SECONDS`: https://github.com/Lilypad-Tech/lilypad/blob/9be369e15b99047b3462b31777882feebbf2c77e/pkg/options/solver.go#L21

### Details 

We mark the deal container with timestamps as the job progresses through its lifecycle. Once the job is complete or times out, we compute job run timings and report them to the stats API.

The solver stats `JobRun` struct includes these new timing fields:

- `MatchDurationMilliseconds`. Time from when a job offer was created until it was matched.
- `ExecutionDurationMilliseconds`. Time from when a match was made until the resource provider uploads job outputs.
- `RetrievalDurationMilliseconds`. Time from when job outputs were uploaded until they are downloaded by the job creator.

We should report values for all three and `TotalDurationMilliseconds` when a job run is successful.

When a job times out, `ExecutionDurationMilliseconds` and `RetrievalDurationMilliseconds` may be zero depending on when the job times out.

We do not report timed out job offers that have not been matched, but may consider adding this reporting in the future.

### Related issues or PRs

Epic: https://github.com/Lilypad-Tech/internal/issues/348


pkg/solver/server.go
```diff
@@ -613,6 +613,12 @@ func (server *solverServer) downloadFiles(res corehttp.ResponseWriter, req *core
 	}
 
 	if err := server.handleFileDownload(GetDealsFilePath(id), res, func() {
+		downloadAt := int(time.Now().UnixNano() / int64(time.Millisecond))
+		deal, err = server.store.UpdateDealDownloadTime(deal.ID, downloadAt)
+		if err != nil {
+			server.log.Error().Err(err).Msg("failed to record deal downloadAt time")
```
narbs91: [nit] Let's also add the `dealId` and `downloadAt` as a Str's here

pkg/solver/server.go
```diff
@@ -749,6 +757,12 @@ func (server *solverServer) uploadFiles(res corehttp.ResponseWriter, req *coreht
 			return err
 		}
 
+		uploadAt := int(time.Now().UnixNano() / int64(time.Millisecond))
+		_, err = server.store.UpdateDealUploadTime(deal.ID, uploadAt)
+		if err != nil {
+			server.log.Error().Err(err).Msg("failed to record deal uploadAt time")
```
narbs91: [nit] Let's also add the `dealId` and `uploadAt` as a Str's here

pkg/solver/server.go
```diff
@@ -825,6 +839,12 @@ func (server *solverServer) jobOfferDownloadFiles(res corehttp.ResponseWriter, r
 	}
 
 	if err := server.handleFileDownload(GetDealsFilePath(jobOffer.DealID), res, func() {
+		downloadAt := int(time.Now().UnixNano() / int64(time.Millisecond))
+		deal, err = server.store.UpdateDealDownloadTime(deal.ID, downloadAt)
+		if err != nil {
+			server.log.Error().Err(err).Msg("failed to record deal downloadAt time")
```
narbs91: [nit] Let's also add the `dealId` and `downloadAt` as a Str's here

narbs91 (commented):
Only small comments on logging but tested locally and looks to be working to me:

image

pkg/solver/server.go
bgins: ✅ Added in 0e00372e06c8e4bb7fd5ea69e28e7e55756e7678

pkg/solver/server.go
bgins: ✅ Added in 0e00372e06c8e4bb7fd5ea69e28e7e55756e7678

pkg/solver/server.go
bgins: ✅ Added in 0e00372e06c8e4bb7fd5ea69e28e7e55756e7678

narbs91 (approved):
LGTM!

PR #563, opened 2025-04-02
feat: Reduce solver control loop triggers

bgins:
### Summary

- [x] Remove solver control loop triggers on web3 events
- [x] Add logging in matcher and store implementations

We would like to reduce the number of events that trigger the solver control loop for a few reasons:

- We have noticed control loop iterations that do not perform useful work (`solve` traces complete in <1ms)
- The web3 event triggers are less relevant with the new protocol
- Rapid control loop iterations may be contributing to controller instability

We also considered removing the control loop trigger when a resource offer is added, but decided against it. This trigger progresses work when there are queued job offers. When a resource offer is posted, we can attempt to match it with a job offer immediately.

We have also added a few logs for better visibility in the store and matcher.

### Test plan

Start the stack. Run a few jobs.


narbs91 (approved):
LGTM!

PR #566, opened 2025-04-14
fix: Prevent duplicate unmatched resource offers

bgins:
### Summary

- [x] Prevent duplicate unmatched resource offers

We've noticed that some resource providers post multiple unmatched resource offers. While we get a handle on the reasons why, we will decline new resource offers when an unmatched resource offer exists.

### Test plan

Start the stack. Run a job.

We haven't been able to replicate the multiple resource offer scenario yet, so this fix is a bit challenging to test. In theory, the resource provider should check if their offers have already been posted:

https://github.com/Lilypad-Tech/lilypad/blob/54076b20be701e0fa8a7ae1a474b73ec79dba9f0/pkg/resourceprovider/controller.go#L268-L274

We can fake a scenario where a resource provider ignores existing resource offers by commenting out this block of code and replacing it with "no active offers":

```go
activeResourceOffers := map[int]data.ResourceOfferContainer{}
```

Run two instances of the resource provider. The first should post an offer. The solver should reject the second offer and log an error message.

### Details

We can't return an error to the resource provider because this error path is fatal and would cause the resource provider to crash. We've left a TODO comment to address this later.

### DIFF
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -449,6 +449,21 @@ func (controller *SolverController) addResourceOffer(resourceOffer data.Resource
 	}
 	resourceOffer.ID = id
 
+	// Check if we have an unmatched resource offer for the resource provider
+	existingOffers, err := controller.store.GetResourceOffers(store.GetResourceOffersQuery{
+		ResourceProvider: resourceOffer.ResourceProvider,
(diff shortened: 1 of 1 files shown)


narbs91 (approved):
LGTM!

image

PR #569, opened 2025-04-16
fix: Target unmatched resource offers

bgins:
### Summary

- [x] Target unmatched resources offers
- [x] Remove `GetResourceOfferByAddress` store method

This pull request fixes an edge case where already matched resource offers could be targeted. The changes query the store for unmatched resource offers, and we make a deal with one if found.

Note that we recently merged changes to prevent duplicate resource offers in #566.

The `GetResourceOfferByAddress` store method was only used for targeting, so we have removed it.

### Test plan

Start the stack and run a job targeting the local development resource provider:

```sh
./stack run cowsay:v0.0.4 -i Message="airy" --target 0x15d34AAf54267DB7D7c367839AAf71A00a2C6A65
```

This should succeed. Now run a job targeting a non-existent resource provider:

```sh
./stack run cowsay:v0.0.4 -i Message="airy" --target 0x11111111111
```

The job should return quickly with a job cancelled error message.

Now run two jobs in quick succession targeting the local development resource provider. While the first job is running, the second job should be rejected because an unmatched resource offer is not available.

### DIFF
--- a/pkg/solver/matcher/matcher.go
+++ b/pkg/solver/matcher/matcher.go
@@ -282,14 +282,14 @@ func getTargetedDeal(
 		trace.WithAttributes(attribute.String("job_offer.target.address", jobOffer.JobOffer.Target.Address)))
 	defer span.End()
 
-	span.AddEvent("db.get_resource_offer_by_address.start")
-	resourceOffer, err := db.GetResourceOfferByAddress(jobOffer.JobOffer.Target.Address)
-	if err != nil {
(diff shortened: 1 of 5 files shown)


hollygrimm (approved):
@bgins the code looks good, I threw 5 targeted jobs at my local RP and there was 1 success, 4 cancelled.

logan-wrld (approved):
Thank you for the walk-through @bgins

PR #570, opened 2025-04-18
feat: Add module metadata fields

bgins:
### Summary

- [x] Add module `author` fields
- [x] Add module `model` fields
- [x] Add module `image` URL field
- [x] Add module `description` field

This pull request extends the module spec to include metadata about the module author, description, image URL and model.

We don't currently use these fields in the core system, but we may want to validate them when loading modules in the future. In addition, the `Module` type is currently our best source of truth for what fields a module contains.


narbs91 (approved):
LGTM!

PR #571, opened 2025-04-21
chore: Update module path casing

bgins:
### Summary

- [x] Update `go.mod` module path casing
- [x] Update build module paths casing
- [x] Update import module paths casing
- [x] Update docs module paths casing

We would like to publish `lilypad` as a Go package to re-use code in Anura and to make it available to other client implementers. In order to publish, we need to update module paths to match the casing in the GitHub URL, which means changing `https://github.com/lilypad-tech/lilypad` to `https://github.com/Lilypad-Tech/lilypad`.

We've also made changes in the docs for consistency.

### Task/Issue reference

Implements: #79 

### Test plan

Start the stack. Run a job or two. Everything should work as expected.

Build a binary with a version and commit SHA including the `v2` path segments:

```sh
go build -o lilypad-test -v -ldflags="-X 'github.com/Lilypad-Tech/lilypad/v2/pkg/system.Version=someversion' -X 'github.com/Lilypad-Tech/lilypad/v2/pkg/system.CommitSHA=somesha'"
```

Run the version command to confirm the version is `someversion` and the commit SHA is `somesha`:

```sh
./lilypad-test version
```

Note that this step confirms the build changes (1440289f0bf16386d72f63cca0c9ac93f3475ccb) will continue to work.


narbs91:
Was able to run a few jobs no problem; however, when I built the binary and tested the version it didn't seem to quite work (at least i don't think)

image

narbs91 (approved):

hollygrimm (approved):
tested as outlined

PR #573, opened 2025-04-22
feat: Add job file inputs

### LINKED ISSUE #572: Add file inputs
## General Description

Lilypad jobs currently accept a text input. We would like to extend jobs to accept file inputs, which will expand the types of jobs we can run.

This functionality affects the solver, resource provider, job creator, and modules.

The module should specify expected input files.

The job creator should validate input files and send them to the solver with the job offer.

The solver should store the input files and allow download from the resource provider who will run the job.

The resource provider should download and validate input files, then provide them to Bacalhau for the job run.

bgins:
### Summary

- [x] Add `inputFiles` field to module spec
- [x] Add `inputFiles` to job creator
  - [x] Add `INPUTS_PATH` config to job creator options
  - [x] Add `inputFiles` to CLI run command
  - [x] Check files in input path match requested `inputFiles`
- [x] Add solver support for job offers with file inputs
  - [x] Add `SERVER_MAX_FILE_INPUTS_MEMORY_MB` to set max memory consumed when processing input files
  - [x] Add `SERVER_MAX_FILE_INPUTS_SIZE_MB` to set max file upload size
  - [x] Add post job offer with files endpoint and handler
  - [x] Add get download inputs file endpoint and handler
  - [x] Add metric to track file input sizes
- [x] Add CLI run command post job offer with files
  - [x] Add solver client and job creator methods to call when `inputFiles` is non-empty 
  - [x]  Add `PostRequestBufferWithHeaders` helper
- [x] Add resource provider support for jobs with file inputs
  - [x] Add download files when `inputFiles`
  - [x] Add Bacalhau support

This pull request adds file intputs to jobs. We want to extend the capability of jobs by allowing audio, video, and other file inputs.

We have also updated the integration tests to create isolated data directories in the GH actions `workspace` for each service.

### Task/Issue reference

Closes: #572 

### Test plan

We have created an example module to demostrate the file inputs functionality: https://github.com/Lilypad-Tech/lilypad-concat-files-module

Rebuild the base services to include the new Bacalhau input sources permission.

```sh
./stack compose-build
```

Create a local `inputs` directory and add `one.txt` and `two.txt` files with some text content in them.

Start the stack. Run a job using the example module.

```sh
./stack run github.com/Lilypad-Tech/lilypad-concat-files-module:main --inputs-path=./inputs
```

Now add `three.txt` and `four.txt` text files and run another job. This confirms optional files will be included and concatenated into the output.

Run another job that places a separator between each of the text file inputs.

```sh
./stack run github.com/Lilypad-Tech/lilypad-concat-files-module:main --inputs-path=./inputs -i Separator="=========="
```

Lastly, check that the max file input size config works by stopping the solver, and running again with a 0MB max input file size.

```sh
./stack solver --server-max-file-inputs-size-mb 0
```

#### Modules

We have added an `inputFiles` field to the module spec. This field declares required and optional input files. For example:

```json
"inputFiles": {
  "required": ["one.txt", "two.txt"],
  "optional": ["three.txt", "four.txt"]
}
```

The `one.txt` and `two.txt` files are required, and `three.txt` and `four.txt` are optional. No other files are allowed.

#### Job creator

We have added file inputs support to the CLI run command. The CLI run command checks for required files and ensures only explicitly declared files are present. All files must be in a single directory without subdirectories.

#### Solver

The solver has two new endpoints:

- `/job_offers/with_files`. Accepts a `multipart/form-data` request with a job offer and a tar archive with input files.
- `/deals/{id}/input_files`. Accepts a request to download the input files tar archive associated with a deal `id` from a resource provider.

A `/job_offers/with_files` route has been added for job creators and Anura. The resource provider is authenticated by signature before they are allowed to download input files.

Uploaded file inputs are stored in a `<DATA_DIR>/job-input-archives/<job-offer-id>` directory (defaults to `/tmp/lilypad/data/job-input-archives/<job-offer-id>`)

#### Resource provider

The resource provider checks the job offer to determine if a job has file inputs. If it does, it requests them from the solver before running the job.

Downloaded file inputs extracted from the tar archive and stored in a `<DATA_DIR>/job-inputs/<deal-id>` directory (defaults to `/tmp/lilypad/data/job-inputs/deal-id>`)

Bacalhau is granted access to the `job-inputs` directory so that it can access file inputs. Note that the current implementation will work for Docker RPs, but may take some extra work for others to configure.

#### Configs

We have added a new `INPUTS_PATH` config to the CLI run command options. An `inputs` directory can be set with a CLI option or an environment variable:

```sh
./stack run github.com/Lilypad-Tech/lilypad-concat-files-module:main --inputs-path=./inputs
INPUTS_PATH=./inputs ./stack run github.com/Lilypad-Tech/lilypad-concat-files-module:main
```

We have also added `SERVER_MAX_FILE_INPUTS_MEMORY_MB` and `SERVER_MAX_FILE_INPUTS_SIZE_MB` configs to the solver server. 

`SERVER_MAX_FILE_INPUTS_MEMORY_MB` sets how much memory is used when accepting file inputs. Larger file inputs can be sent, but anything in excess of the memory max is temporarily stored on disk.

```sh
./stack solver --server-max-file-inputs-memory-mb 10
SERVER_MAX_FILE_INPUTS_MEMORY_MB=10 ./stack solver
```

`SERVER_MAX_FILE_INPUTS_SIZE_MB` sets the maxiumum uploaded input files size.

```sh
./stack solver --server-max-file-inputs-size-mb 10
SERVER_MAX_FILE_INPUTS_SIZE_MB=10 ./stack solver
```

#### Metrics

We have added metrics to track job offers with and without file inputs, and the size of file inputs when present. In our Grafana dashboard, we plan to display percent of jobs with file inputs and average file inputs size:

CleanShot 2025-05-05 at 14 36 58@2x

### Related issues or PRs

Design doc: https://www.notion.so/lilypadnetwork/File-Inputs-1d1155da99b5806ea3c8fa0f7788e7ae
Observability PR: https://github.com/Lilypad-Tech/observability/pull/33

### DIFF
--- a/pkg/data/data.go
+++ b/pkg/data/data.go
@@ -45,6 +45,9 @@ type Module struct {
 	Image string      `json:"image"`
 	Model ModuleModel `json:"model"`
 
+	// Expected input files
+	InputFiles InputFiles `json:"inputFiles"`
+
 	// the min spec that this module requires
 	// e.g. does this module need a GPU?
 	// the module file itself will contain this spec
@@ -68,6 +71,11 @@ type ModuleModel struct {
 	Name string `json:"name"`
 	Type string `json:"type"`
 	Size string `json:"size"`
+}
+
+type InputFiles struct {
+	Required []string `json:"required"`
+	Optional []string `json:"optional"`
 }
 
 // describes a workload to be run
--- a/pkg/data/data.go
+++ b/pkg/data/data.go
@@ -207,6 +207,8 @@ type JobOffer struct {
 	// the user inputs to the module
 	// these values will power the go template
 	Inputs map[string]string `json:"inputs"`
+	// Expected input files
+	InputFiles InputFiles `json:"inputFiles"`
 	// tells the solver how to match these prices
 	// for JC this will normally be MarketPrice
 	Mode PricingMode `json:"mode"`
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -526,13 +526,13 @@ func PostRequestBufferWithHeaders[ResultType any](
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
 		log.Debug().Msgf("error reading response body: %s", body)
-		return result, err
+		return result, fmt.Errorf("error reading response body: %s", body)
 	}
 
 	err = json.Unmarshal(body, &result)
 	if err != nil {
 		log.Debug().Msgf("error unmarshaling response body: %s", body)
-		return result, err
+		return result, fmt.Errorf("error unmarshaling response body: %s", body)
 	}
 
 	return result, nil
--- a/pkg/http/utils.go
+++ b/pkg/http/utils.go
@@ -494,6 +494,50 @@ func PostRequestBuffer[ResultType any](
 	return result, nil
 }
 
+func PostRequestBufferWithHeaders[ResultType any](
+	options ClientOptions,
+	path string,
+	headers map[string]string,
+	data *bytes.Buffer,
+) (ResultType, error) {
+	var result ResultType
+	client := newRetryClient()
+	privateKey, err := web3.ParsePrivateKey(options.PrivateKey)
+	if err != nil {
+		return result, err
+	}
+	req, err := retryablehttp.NewRequest("POST", URL(options, path), data)
+	if err != nil {
+		return result, err
+	}
+
+	if err := AddHeaders(req, privateKey, web3.GetAddress(privateKey).String()); err != nil {
+		return result, err
+	}
+	for key, value := range headers {
+		req.Header.Set(key, value)
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return result, err
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Debug().Msgf("error reading response body: %s", body)
+		return result, err
+	}
+
+	err = json.Unmarshal(body, &result)
+	if err != nil {
+		log.Debug().Msgf("error unmarshaling response body: %s", body)
+		return result, err
+	}
+
+	return result, nil
+}
+
 func newRetryClient() *retryablehttp.Client {
 	retryClient := retryablehttp.NewClient()
 	retryClient.RetryMax = 10
--- a/pkg/jobcreator/controller.go
+++ b/pkg/jobcreator/controller.go
@@ -116,6 +116,7 @@ func (controller *JobCreatorController) AddJobOfferWithFiles(offer data.JobOffer
 	controller.log.Debug("add job offer with files", offer)
 	container, err := controller.solverClient.AddJobOfferWithFiles(offer, inputsPath)
 	if err != nil {
+		controller.log.Error("unable to post job with files", err)
 		return container, err
 	}
 
--- a/pkg/jobcreator/controller.go
+++ b/pkg/jobcreator/controller.go
@@ -112,6 +112,22 @@ func (controller *JobCreatorController) AddJobOffer(offer data.JobOffer) (data.J
 	return container, nil
 }
 
+func (controller *JobCreatorController) AddJobOfferWithFiles(offer data.JobOffer, inputsPath string) (data.JobOfferContainer, error) {
+	controller.log.Debug("add job offer with files", offer)
+	container, err := controller.solverClient.AddJobOfferWithFiles(offer, inputsPath)
+	if err != nil {
+		return container, err
+	}
+
+	// Set the controller's jobID to track this specific job
+	controller.jobID = container.JobOffer.ID
+	controller.log.Debug("Set controller jobID", map[string]any{
+		"jobID": controller.jobID,
+	})
+
+	return container, nil
+}
+
 // SubscribeToJobOfferUpdates subscribes to job offer updates for this controller's job
 // Returns a function that can be called to unsubscribe
 func (controller *JobCreatorController) SubscribeToJobOfferUpdates(sub JobOfferSubscriber) func() {
--- a/pkg/jobcreator/utils.go
+++ b/pkg/jobcreator/utils.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Lilypad-Tech/lilypad/v2/pkg/data"
 	"github.com/Lilypad-Tech/lilypad/v2/pkg/module"
+	"github.com/rs/zerolog/log"
 
 	nanoid "github.com/matoous/go-nanoid/v2"
 )
@@ -23,6 +24,10 @@ func getJobOfferFromOptions(options JobCreatorOfferOptions, jobCreatorAddress st
 	if module.HasInputFiles(loadedModule.InputFiles) {
 		err = module.ValidateInputFiles(options.InputsPath, loadedModule.InputFiles)
 		if err != nil {
+			log.Error().Err(err).
+				Str("inputsPath", options.InputsPath).
+				Any("inputFiles", loadedModule.InputFiles).
+				Msg("failed to validate input files")
 			return data.JobOffer{}, fmt.Errorf("error reading input files: %s", err.Error())
 		}
 	}
--- a/pkg/jobcreator/utils.go
+++ b/pkg/jobcreator/utils.go
@@ -20,6 +20,13 @@ func getJobOfferFromOptions(options JobCreatorOfferOptions, jobCreatorAddress st
 		return data.JobOffer{}, fmt.Errorf("error loading module: %s opts=%+v", err.Error(), options)
 	}
 
+	if module.HasInputFiles(loadedModule.InputFiles) {
+		err = module.ValidateInputFiles(options.InputsPath, loadedModule.InputFiles)
+		if err != nil {
+			return data.JobOffer{}, fmt.Errorf("error reading input files: %s", err.Error())
+		}
+	}
+
 	// Generate a nonce to make sure the job offer is unique
 	nonce, err := nanoid.New()
 	if err != nil {
@@ -34,6 +41,7 @@ func getJobOfferFromOptions(options JobCreatorOfferOptions, jobCreatorAddress st
 		Module:     options.Module,
 		Spec:       loadedModule.Machine,
 		Inputs:     options.Inputs,
+		InputFiles: loadedModule.InputFiles,
 		Mode:       options.Mode,
 		Pricing:    options.Pricing,
 		Timeouts:   options.Timeouts,
--- a/pkg/module/utils.go
+++ b/pkg/module/utils.go
@@ -1,10 +1,12 @@
 package module
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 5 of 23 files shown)


pkg/http/utils.go
```diff
@@ -494,6 +494,50 @@ func PostRequestBuffer[ResultType any](
...
 }
 
+func PostRequestBufferWithHeaders[ResultType any](
+	options ClientOptions,
+	path string,
+	headers map[string]string,
+	data *bytes.Buffer,
+) (ResultType, error) {
+	var result ResultType
+	client := newRetryClient()
```
narbs91, line 504: Is there a way to hoist this up to the module level so that we don't have to create one on every call to `PostRequestBufferWithHeaders`?

pkg/jobcreator/controller.go
```diff
@@ -112,6 +112,22 @@ func (controller *JobCreatorController) AddJobOffer(offer data.JobOffer) (data.J
 	return container, nil
 }
 
+func (controller *JobCreatorController) AddJobOfferWithFiles(offer data.JobOffer, inputsPath string) (data.JobOfferContainer, error) {
+	controller.log.Debug("add job offer with files", offer)
+	container, err := controller.solverClient.AddJobOfferWithFiles(offer, inputsPath)
+	if err != nil {
+		return container, err
```
narbs91, line 120: Let's add an error log here

pkg/jobcreator/utils.go
```diff
@@ -20,6 +20,13 @@ func getJobOfferFromOptions(options JobCreatorOfferOptions, jobCreatorAddress st
 		return data.JobOffer{}, fmt.Errorf("error loading module: %s opts=%+v", err.Error(), options)
 	}
 
+	if module.HasInputFiles(loadedModule.InputFiles) {
+		err = module.ValidateInputFiles(options.InputsPath, loadedModule.InputFiles)
+		if err != nil {
+			return data.JobOffer{}, fmt.Errorf("error reading input files: %s", err.Error())
```
narbs91, line 31: let's add an error log here

pkg/module/utils.go
```diff
@@ -843,3 +844,51 @@ func getHashBasedRepoPath(repoURL, hash string) (string, error) {
...
+// Check that required files exist and only explicitly defined
+// files are in the target directory
+func ValidateInputFiles(path string, inputFiles data.InputFiles) error {
+	allowedFiles := make(map[string]bool)
+	for _, file := range inputFiles.Required {
+		allowedFiles[file] = true
+	}
+	for _, file := range inputFiles.Optional {
+		allowedFiles[file] = true
+	}
```
narbs91: Is it possible for the file names to clash? Like if someone were to have the same file name in both optional and required, I think this would overwrite the map value.  Don't know why someone might do that but an edge case to consisder

pkg/data/data.go
```diff
@@ -199,6 +207,8 @@ type JobOffer struct {
 	// the user inputs to the module
 	// these values will power the go template
 	Inputs map[string]string `json:"inputs"`
+	// Expected input files
+	InputFiles InputFiles `json:"inputFiles"`
```
narbs91, line 211: Have you tested this with Anura? Theoretically this should not break Anura because it's making its' own module and then passing it to the solver; however, doesn't hurt to double check running Anura against this version of the solver locally just to make sure nothing breaks

pkg/options/job-creator.go
```diff
@@ -84,6 +87,16 @@ func CheckJobCreatorOptions(options jobcreator.JobCreatorOptions) error {
 		return err
 	}
 
+	if options.Offer.InputsPath != "" {
+		fileInfo, err := os.Stat(options.Offer.InputsPath)
+		if err != nil {
+			return fmt.Errorf("inputs path error: %w", err)
+		}
+		if !fileInfo.IsDir() {
+			return fmt.Errorf("inputs path must be a directory: %s", options.Offer.InputsPath)
```
narbs91, line 96: Maybe we add some logging here unless you already have it upstream

pkg/options/server.go
```diff
@@ -38,6 +39,13 @@ func GetDefaultRateLimiterOptions() http.RateLimiterOptions {
 	}
 }
 
+func GetDefaultStorageOptions() http.StorageOptions {
+	return http.StorageOptions{
+		MaximumFileInputsMemoryMB: GetDefaultServeOptionInt("SERVER_MAX_FILE_INPUTS_MEMORY_MB", 1),
+		MaximumFileInputsSizeMB:   GetDefaultServeOptionInt("SERVER_MAX_FILE_INPUTS_SIZE_MB", 10),
```
narbs91, line 45: Just something to consider when we roll this out but we should probs up the Solver disk space on the machine

pkg/solver/server.go
```diff
@@ -376,6 +379,145 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
...
+	signerAddress, err := http.CheckSignature(req)
+	if err != nil {
+		server.log.Error().Err(err).Msg("error checking signature")
+		corehttp.Error(res, "Unauthorized", corehttp.StatusUnauthorized)
+		return
+	}
+
+	// Set a maximum size for the entire request body
+	// This includes the job offer, which only accounts for a small portion of the request body
+	maxRequestSize := int64(server.options.Storage.MaximumFileInputsSizeMB << 20)
```
narbs91: [nit] maybe rename to maxRequestSizeInBytes?

pkg/solver/server.go
```diff
@@ -376,6 +379,145 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
...
+	req.Body = corehttp.MaxBytesReader(res, req.Body, maxRequestSize)
+
+	// Parse the multipart form with max memory size
+	err = req.ParseMultipartForm(int64(server.options.Storage.MaximumFileInputsMemoryMB << 20))
```
narbs91, line 406: Looks like you have `int64(server.options.Storage.MaximumFileInputsMemoryMB << 20)` computed already with `maxRequestSize` so you can probs re-use that unless i've missed a detail

pkg/solver/server.go
```diff
@@ -376,6 +379,145 @@ func (server *solverServer) addJobOffer(jobOffer data.JobOffer, res corehttp.Res
...
+	// Check version
+	if server.options.AccessControl.EnableVersionCheck && !http.IsAnura(req, server.options.AccessControl.AnuraAddresses) {
+		versionHeader, _ := http.GetVersionFromHeaders(req)
+		minVersion, ok := server.versionConfig.IsSupported(versionHeader)
+		if !ok {
+			server.log.Debug().Str("cid", jobOffer.ID).
+				Str("address", jobOffer.JobCreator).
+				Str("version", versionHeader).
+				Str("minVersion", minVersion).
+				Msg("job offer rejected because job creator is running an unsupported version")
```
narbs91:
```suggestion
				Msg("job offer with files rejected because job creator is running an unsupported version")
```

pkg/module/utils.go
```diff
@@ -843,3 +844,51 @@ func getHashBasedRepoPath(repoURL, hash string) (string, error) {
...
+
+// File inputs
+
+func HasInputFiles(inputFiles data.InputFiles) bool {
+	return len(inputFiles.Required) > 0 || len(inputFiles.Optional) > 0
+}
+
+// Check that required files exist and only explicitly defined
+// files are in the target directory
+func ValidateInputFiles(path string, inputFiles data.InputFiles) error {
```
narbs91, line 859: What do you think about having an additional check inside this function to prevent like .exe and other malicious things from coming in?

We could come up with a list of allowed types or a list of disallowed types depending on which list will be shorter

narbs91 (commented):
Looking great so far, left some initial comments

pkg/http/utils.go
bgins: Yeah, I could see adding a `HTTPClient` with a retry client on it. Then convert this function and the others to methods on it.

Would be a bit of refactoring work, would propose we look at it in a separate PR.

pkg/module/utils.go
bgins: We could, but I'm not sure how much it improves our security.

Consider this scenario. A module author specifies a required `mybad.txt` file, then once the file is available to the module code, they rename it to `mybad.exe`.

pkg/module/utils.go
bgins: That's a great point. Lemme add something to validate no duplicate filenames or the job fails.

pkg/options/server.go
bgins: Yep 💯

pkg/solver/server.go
bgins: This one uses the `MaximumFileInputsMemoryMB` config, where `maxRequestSize` uses `MaximumFileInputsSizeMB` (names are quite close, `Memory` vs `Size`).

pkg/module/utils.go
narbs91: Ah gotcha, yeah, I wonder if its just a matter of adding a check in the RP side down the line to have them have a file check or something to filter out file formats to exclude from processing

pkg/options/job-creator.go
bgins: These errors will be fatal and dumped when the program exits here:

https://github.com/Lilypad-Tech/lilypad/blob/62253a1bc79d81d87212a7932f52ee79de3736dd/cmd/lilypad/run.go#L33-L36

That seem alright?

pkg/jobcreator/controller.go
bgins: ✅ Done in 9152933

pkg/jobcreator/utils.go
bgins: ✅ Done in 9152933

pkg/module/utils.go
bgins: Yeah, we could do some analysis on the files. Go deeper than just validating file names. Do you have ideas for how we would approach this?

narbs91: Hmm not off the top of my head but its something we can brainstorm on together, i'm sure there's a way

pkg/solver/server.go
narbs91: Ahh yeah, there indeed was a detail i missed 😆

pkg/options/job-creator.go
narbs91: Sure, let's go with that and then fix if/when we need to

pkg/solver/server.go
bgins: ✅ Done in ba51a97

pkg/module/utils.go
bgins: ✅ Done in fc754a373b8db8d726e341b6e3402e38a2be5e52

pkg/module/utils.go
bgins: Added some basic checks here: https://github.com/Lilypad-Tech/lilypad/pull/573/commits/1b074d1f8eea20a3ce21dbd8ada315676f2c94a1

narbs91 (approved):
LGTM!

walkah (approved):
Was able to successfully test on a GPU RP - works well! 🚀

PR #579, opened 2025-05-14
fix: version ARGs for RP Dockerfile

walkah:
### Summary

This PR properly scopes the `VERSION` and `COMMIT_SHA` args so that docker builds of the RP properly report this information. These values are set in the `release_docker.yml` workflow: https://github.com/Lilypad-Tech/lilypad/blob/main/.github/workflows/release_docker.yml#L44-L46 however the args weren't properly scoped to the actual build stages in the Dockerfile.


bgins (approved):
Looks good! 🙌

Issue #493, opened 2025-01-17
Improve resource provider module allowlist parsing


bgins:
## General Description

A resource provider can specify which modules they are willing to run. Unfortunately, they have to specify the module CID in the module allowlist for the solver matching comparison to work correctly.

For example, one could start a resource provider:

```sh
./stack resource-provider --offer-modules "QmVvTRUBJiiUmEp3P1ubU3djgsbABdF3xswouVpM21rCto"
```

`QmVvTRUBJiiUmEp3P1ubU3djgsbABdF3xswouVpM21rCto` is the module CID for `cowsay:v0.0.4`. One could run a job against this resource provider with:

```sh
./stack run cowsay:v0.0.4 -i Message="cids"
```

It's unrealistic to expect either resource providers or module authors to compute these CIDs. It would be better if the resource provider could specify modules in the same way as the CLI run command.

## Which system(s) or functionality does this affect

These changes will affect the solver, resource provider, or both depending on how we implement a fix.

## Describe the changes, and how this affects/ interacts with each system.

The necessary changes are TBD.

PR #581, opened 2025-05-27
perf: Remove supported modules check from solver matching algorithm

bgins:
### Summary

- [x] Remove supported modules check from matching algo
  - [x] Remove `moduleIDError` match result
  - [x] Remove `moduleMismatch` match result
  - [x] Remove module ID dependeny from `priceMismatch` match result
  - [x] Update tests to remove supported module test cases
- [x] Prevent resource providers from using `OFFER_MODULES` config 

The solver matching algorithm is underperforming and we want it to run faster.

### Test plan

Run the benchmarks on `main`:

```sh
./stack benchmarks-solver
```

Run them again on this branch to observe the improved results.

Start a resource provider with a module list:

```
./stack resource-provider --offer-modules "QmVvTRUBJiiUmEp3P1ubU3djgsbABdF3xswouVpM21rCto"
OFFER_MODULES=QmVvTRUBJiiUmEp3P1ubU3djgsbABdF3xswouVpM21rCto ./stack resource-provider
```

Both commands should fail with a `Error: OFFER_MODULES module list is not currently supported` error message.

Lastly, start the stack. Run a few jobs. Make sure everything generally works.

### Details

The resource offer has a list of supported modules that must be expressed as CIDs. As we have discussed, this mechanism is inconvenient for resource providers, and we could use the module identifier instead: https://github.com/Lilypad-Tech/lilypad/issues/493

We are not aware of any resource providers that use the module list mechanism, so we have disabled it and will revisit in future work.

The solver matching algorithm compares the requested job offer module against the resource provider module list. The job offer does not have a module CID, so we compute it on the fly. This check is the main bottleneck in the solver matching algorithm, and because it is unused, we remove it in this pull request.

### DIFF
--- a/pkg/options/resource-provider.go
+++ b/pkg/options/resource-provider.go
@@ -147,6 +147,10 @@ func CheckResourceProviderOfferOptions(options resourceprovider.ResourceProvider
 		return fmt.Errorf("OFFER_RAM cannot be zero")
 	}
 
+	if len(options.Modules) > 0 {
+		return fmt.Errorf("OFFER_MODULES module list is not currently supported")
+	}
+
 	return nil
 }
 
--- a/pkg/solver/matcher/match.go
+++ b/pkg/solver/matcher/match.go
@@ -144,7 +144,6 @@ func (result marketPriceUnavailable) attributes() []attribute.KeyValue {
 type priceMismatch struct {
 	resourceOffer data.ResourceOffer
 	jobOffer      data.JobOffer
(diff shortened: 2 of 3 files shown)


narbs91 (approved):
LGTM! Was able to confirm the module list indeed does return an unsupported error on startup and was able to run a couple of cowsay jobs. Looks like the integration tests are failing though (might just be a flaky test run)

Issue #559, opened 2025-03-28
Replace `AgreementState` with new job states


bgins:
## General Description

We currently track job states with `AgreementState`: https://github.com/Lilypad-Tech/lilypad/blob/50788836d22f70b134acfa88225e8fb3ea1bbd35/pkg/data/enums.go#L18-L31

We would like to update the job states to reflect the new protocol and other changes that we are making across the system. In addition, we would like to move away from using an enum for job state. It's inconvenient and inflexible.

### New job states

An initial sketch of job states we have discussed:

- **Unmatched.** A job offer or resource offer that hasn't been matched.
- **Matching.** A job offer or resource offer currently being matched.
- **DealCreated.** A match was made, and the job is ready for execution or a resource provider is actively running it.
- **ResultsSubmitted.** The resource provider has completed the job and uploaded the job outputs to the solver.
- **ResultsAccepted.** The job creator has downloaded the job outputs. This state indicates a successful job run.
- **ResultsRejected.** The job creator has rejected the job results and requested validation.
- **ValidationPending.** Validation has been requested and is in progress.
- **ValidationAccepted.** The job run was validated as correct.
- **ValidationRejected.** The job run was rejected as incorrect.
- **TimeoutMatch.** The job timed out before it could be matched.
- **TimeoutExecution.** The job timed out while the resource provider was executing it (any time before job output upload)
- **TimeoutDownload.** The job timed out waiting for the job creator to download job outputs from the solver.
- **TimeoutValidation.** The validation timed out.
- **JobCancelled.** The job was cancelled.

🗒 **Please edit these states directly as we evolve our plan for them.**

### Implementation

We will implement these changes behind our `lilynext` feature flag. We plan to take on the work alongside implementing the new protocol.

## Which system(s) or functionality does this affect

This functionality affects the solver, job creator, resource provider, protocol-smart-contracts, apis, and observability.

## Describe the changes, and how this affects/ interacts with each system

The changes interact with many parts of the system:

- **Solver.** Sets and reads states throughout the job lifecycle.
- **Job creator.** Downloads job outputs when results have been submitted.
- **Resource Provider.** Receives jobs on match and submits job ouputs.
- **Protocol.** Some states are tracked in the protocol.
- **APIs.** The stats API receives job run stats with these states.
- **Observability.** Metrics about job disposition are sent to the observability stack.

PR #585, opened 2025-05-29
feat: Add job lifecycle timeouts

### LINKED ISSUE #560: Add job lifecycle timeouts
## General Description

We recently added a global job timeout in #511. This timeout is a blunt instrument to clear out stuck and abandoned jobs, but it is not fine-grained enough to time out with purpose.

Let's add the following timeouts to our system:

- **Match timeout.** The maximum time to match a job offer.
- **Execution timeout.** The maximum time a resource provider has to run a module. Starts when the deal is made and ends when the resource provider has uploaded the job outputs.
- **Download timeout.** The maximum time a job creator has to download job outputs.

The solver sets the match and download timeouts as global configs. 

The execution timeout will be set per module to account for variable execution times. The module author and resource provider set this timeout.

The solver tracks timeouts and coordinates cancellation when a timeout occurs.

This functionality affects the solver, job creator and resource provider. 

The solver will implement a global match timeout. The job creator is notified on match timeout with a "network too busy" error message.

The solver will track an execution timeout starting when the deal is made and ending when job outputs are uploaded. The resource provider and job creator are notified on execution timeout with a "execution timed out" error message.

The solver will implement a global download timeout. After execution, the job creator must download job outputs before the download timeout. The job outputs are deleted after the timeout.

If a job output upload or download are in progress when a timeout occurs, they are allowed to continue. Reliable delivery is more important than strict deadlines.

bgins:
### Summary

- [x] Add job lifecycle timeouts 
  - [x] Add match timeout 
  - [x] Add execution timeout
    - [x] Add module spec field to set execution timeout 
  - [x] Add download timeout 
- [x] Remove old timeout fields and metrics tracking
- [x] Fix a small naming type in the resource provider

This pull request adds job lifecycle timeouts for fine-grained control over when a job times out.

This work largely supercedes the global timeout added in #512, but we have kept to global timeout as a fallback if one of the other timeouts fails.

### Task/Issue reference

Closes: #560

### Test plan

Start the base services and the stats API. We will use the stats API to confirm each of the timeouts has occured.

First, start the solver using the default timeouts and run a cowsay job. Check that stats API database and you should see match, execution, retrieval, and total times reported.

```
```

Test the match timeout with a zero-second timeout. 
```
./stack solver --enable-stats true --stats-url http://localhost:8080 --timeout-match-seconds 0
```
The job should time out with a "Job timed before a match could be made" error message. This test case will not report a job run to the stats API, because we only report when a deal has been made.

Test the fallback execution timeout with a zero-second timeout.
```
./stack solver --enable-stats true --stats-url http://localhost:8080 --timeout-execution-seconds 0
```
The job should fail with a "Job timed during execution" error. The stats API should show a match time, but 0 for the execution and retrieval values.

Test the retrieval timeout with a zero-second timeout.
```
./stack solver --enable-stats true --stats-url http://localhost:8080 --timeout-download-seconds 0
```
The job should fail with a "Job output download window expired." error. The stats API should show match and execution times, but 0 for download time.

Lastly, test a job that uses a module-defined execution timeout. We won't time this one out immediately, but we want to observe that the module timeout overrides the solver fallback timeout.

Print `executionTime` immediately after it is set: https://github.com/Lilypad-Tech/lilypad/blob/0763f797c4f7c042456999cb0c15315e7a194f1f/pkg/solver/controller.go#L388-L391

Run a cowsay with on the following commit that sets the execution timeout to 300 seconds.
```
./stack run github.com/Lilypad-Tech/lilypad-module-cowsay:5d7dc7eba59698721e1b1b0aefd72caf2cfaaf0f -i Message="Strawberries"
```

This timeout should override the solver default 900 seconds. The run should succeed and report to the stats API.

#### Configs

We have added configs to set the solver match, execution, download, and overall timeouts. The execution timeout is only applied when the module does not specify one (which fills in as `0` on the module).

Start the solver with a 60 second `TIMEOUT_MATCH_SECONDS`.
```
./stack solver --timeout-match-seconds 60
TIMEOUT_MATCH_SECONDS=60 ./stack solver
```

Start the solver with a fallback 10-minute `TIMEOUT_EXECUTION_SECONDS`. 
```
./stack solver --timeout-execution-seconds 600
TIMEOUT_EXECUTION_SECONDS=600 ./stack solver
```

Start the solver with a two hour `TIMEOUT_DOWNLOAD_SECONDS`.
```
./stack solver --timeout-download-seconds 7200
TIMEOUT_DOWNLOAD_SECONDS=7200 ./stack solver
```

Start the solver with a 12-hour global `TIMEOUT_TOTAL_SECONDS`.
```
./stack solver --timeout-total-seconds 43200
TIMEOUT_TOTAL_SECONDS=43200 ./stack solver
```

#### Module-defined execution timeout

Module authors can set an execution time on their module. We expect they will gather initial data on execution times through our upcoming module benchmarking process.

An example execution timeout can be seen here: https://github.com/Lilypad-Tech/lilypad-module-cowsay/blob/5d7dc7eba59698721e1b1b0aefd72caf2cfaaf0f/lilypad_module.json.tmpl#L7

When the module does not set an execution timeout, we default to the fallback timeout set by the solver.

#### Timeouts

We track four time span in our system:
  - Match time. The time a job offer can wait for a match before the solver times it out.
  - Execution time. The time from when the deal is made until the resource provider uploads the job outputs to the solver.
  - Download time. The time the solver will store job outputs for the job creator to download.
  - Total job run time. The time from when the job offer is created until the job creator downloads the job outputs.
    Includes match, execution, and download times.

We set timeouts for each of these time spans. The module author sets the execution timeout. The solver sets match, download, and total timeouts, where the total timeout is a fallback in case one of the other time outs did not occur. 

The solver also sets a fallback execution timeout for modules that have not specified a timeout.

We have added job states to account for jobs that have timed and when:

https://github.com/Lilypad-Tech/lilypad/blob/0763f797c4f7c042456999cb0c15315e7a194f1f/pkg/data/enums.go#L35-L37

We have also added `Uploading` and `Downloading` states:

https://github.com/Lilypad-Tech/lilypad/blob/0763f797c4f7c042456999cb0c15315e7a194f1f/pkg/data/enums.go#L33-L34

We mark jobs with these states to prevent them from timing out. It doesn't make sense to cancel if the work is done, and we're almost there!

#### Old timeouts

We have removed the old timeout configs and tracking. It looks like they were laying the foundation to fill out the initial vision, but they weren't really used.

We left a set of hardcoded values to fulfill the old protocol expectation, because we weren't certain these could safely be removed. Once we have moved to the new protocol, we can remove them.

We also left the old timeouts in the `AgreementState` enum. We didn't want to disrupt any enum ordering expecations that may exist here or in integrations. We will eventually remove the enum for another mechanism: https://github.com/Lilypad-Tech/lilypad/issues/559

### Related issues or PRs

Issue: #560 
Notion doc: https://www.notion.so/lilypadnetwork/Job-Failures-Timeouts-174155da99b58082a63ac4f8eed2b27d?source=copy_link#174155da99b5804cb7e4caaadf1f4f6c
Module-defined execution timeout example: https://github.com/Lilypad-Tech/lilypad-module-cowsay/pull/4


cmd/lilypad/run.go
```diff
@@ -123,6 +123,15 @@ func runJob(cmd *cobra.Command, options jobcreator.JobCreatorOptions, network st
 		case "JobOfferCancelled":
 			desc = "Job cancelled..."
 			emoji = "😭"
+		case "TimeoutMatch":
+			desc = "Job timed before a match could be made..."
```
narbs91:
```suggestion
			desc = "Job timed out before a match could be made..."
```

cmd/lilypad/run.go
```diff
@@ -123,6 +123,15 @@ func runJob(cmd *cobra.Command, options jobcreator.JobCreatorOptions, network st
...
+			emoji = "⏱️"
+		case "TimeoutExecution":
+			desc = "Job timed during execution..."
```
narbs91:
```suggestion
			desc = "Job timed out during execution..."
```

pkg/options/timeout.go
```diff
@@ -1,62 +1,34 @@
...
-		MediateResults: data.DealTimeout{
-			Timeout:    GetDefaultServeOptionUint64("TIMEOUT_MEDIATE_RESULTS_COLLATERAL", 3600),
-			Collateral: GetDefaultServeOptionUint64("TIMEOUT_MEDIATE_RESULTS_COLLATERAL", 1),
-		},
+func GetDefaultTimeoutOptions() solver.SolverTimeoutOptions {
+	return solver.SolverTimeoutOptions{
+		MatchSeconds:     GetDefaultServeOptionUint64("TIMEOUT_MATCH_SECONDS", 30),
+		ExecutionSeconds: GetDefaultServeOptionUint64("TIMEOUT_EXECUTION_SECONDS", 900), // 15 minutes
+		DownloadSeconds:  GetDefaultServeOptionUint64("TIMEOUT_DOWNLOAD_SECONDS", 3600), // One hour
+		TotalSeconds:     GetDefaultServeOptionUint64("TIMEOUT_TOTAL_SECONDS", 86400),   // One day
```
narbs91: Maybe we want to change the TotalSecond default to just be the sum of the default match, execution and download. One day seems like a lot 😅

pkg/solver/server.go
```diff
@@ -851,6 +853,17 @@ func (server *solverServer) downloadFiles(res corehttp.ResponseWriter, req *core
 		}
 	}
 
+	// The job creator has been verified, mark and start the download.
+	// We mark a Downloading state to avoid timeouts once the download has begun.
+	deal, err = server.store.UpdateDealState(deal.ID, data.GetAgreementStateIndex("Downloading"))
+	if err != nil {
+		server.log.Error().Str("cid", id).Msg("unable to update deal to downloading state")
```
narbs91: Let's add the `Deal.ID` to the log as well

pkg/solver/server.go
```diff
@@ -965,6 +980,13 @@ func (server *solverServer) uploadFiles(res corehttp.ResponseWriter, req *coreht
 			return fmt.Errorf("resource provider address does not match signer address")
 		}
 
+		// The resource provider has been verified, mark and start the upload
+		deal, err = server.store.UpdateDealState(deal.ID, data.GetAgreementStateIndex("Uploading"))
+		if err != nil {
+			server.log.Error().Str("cid", id).Msg("unable to update deal to uploading state")
```
narbs91: Same comment about the `Deal.ID` here

pkg/solver/controller.go
```diff
@@ -403,6 +429,62 @@ func (controller *SolverController) cancelExpiredJobs(ctx context.Context) error
...
+			span.RecordError(err)
+		}
+
+		return jobOffer.ID, ""
+	}
+
+	// Cancel expired job offer, resource offer, and deal
+	deal, err := controller.updateDealState(jobOffer.DealID, data.GetAgreementStateIndex(expirationState))
+	if err != nil {
+		controller.log.Error().Func(system.AddTraceContext(span)).Err(err).Msg("update expired deal state failed")
```
narbs91: Let's add some contextual information from the JobOffer in here to help with debugging

pkg/solver/controller.go
```diff
@@ -403,6 +429,62 @@ func (controller *SolverController) cancelExpiredJobs(ctx context.Context) error
...
+		trace.WithAttributes(attribute.String("job_offer.state", data.GetAgreementStateString(jobOffer.State))),
+		trace.WithAttributes(attribute.String("expiration_state", expirationState)),
+	)
+	defer span.End()
+
+	// Cancel pre-match expired job offer
+	if jobOffer.DealID == "" {
+		_, err := controller.updateJobOfferState(jobOffer.ID, jobOffer.DealID, data.GetAgreementStateIndex(expirationState))
+		if err != nil {
+			controller.log.Error().Func(system.AddTraceContext(span)).Err(err).Msg("update expired job offer state failed")
```
narbs91: Let's add some contextual information from the JobOffer in here to help with debugging

narbs91 (commented):
Looking great overall, just a handful of small comments

cmd/lilypad/run.go
bgins: Fixed in 797320ccec82b1431ffe149c215f23435506c4fd ✅

cmd/lilypad/run.go
bgins: Fixed in 797320ccec82b1431ffe149c215f23435506c4fd ✅

pkg/solver/server.go
bgins: Fixed in b42b6f81a7a53c2f4240daa308e0064bb1517056 ✅

pkg/solver/server.go
bgins: Fixed in b42b6f81a7a53c2f4240daa308e0064bb1517056 ✅

pkg/solver/controller.go
bgins: Fixed in b42b6f81a7a53c2f4240daa308e0064bb1517056 ✅

pkg/solver/controller.go
bgins: Fixed in b42b6f81a7a53c2f4240daa308e0064bb1517056 ✅

narbs91 (approved):
LGTM!

PR #591, opened 2025-07-15
feat: Add reputation tracking for job lifecycle events

rexdotsh:
### Summary

- [x] Add reputation tracking for successful job matches
- [x] Add reputation tracking for failed jobs due to timeouts
- [x] Add reputation tracking for jobs rejected during validation

This PR implements reputation tracking for resource providers across the complete job lifecycle by adding calls to previously unused reputation builder methods (`WithJobMatched`, `WithJobFailed`, `WithValidationLost`).

### Details

The reputation tracking system had several builder methods that were unused:
- `WithJobMatched` - was not being called anywhere
- `WithJobFailed` - was not being called anywhere  
- `WithValidationLost` - was not being called anywhere

This PR adds the missing calls at the appropriate points in the job lifecycle:
- `WithJobMatched(true)` in `addDeal()` - when deals transition to `DealAgreed` state
- `WithJobFailed(true)` in `cancelExpiredJob()` - for all timeout scenarios (`TimeoutMatch`, `TimeoutExecution`, `TimeoutDownload`, `JobTimedOut`)
- `WithValidationLost(true)` in `updateDealState()` - for `MediationRejected` and `ResultsRejected` states

### DIFF
--- a/pkg/solver/controller.go
+++ b/pkg/solver/controller.go
@@ -469,6 +469,11 @@ func (controller *SolverController) cancelExpiredJob(ctx context.Context, jobOff
 
 	// Report expired job
 	controller.stats.PostJobRun(deal)
+	controller.stats.PostReputation(deal.ResourceProvider,
+		stats.NewReputationBuilder().
+			WithJobFailed(true).
+			Build(),
+	)
 
 	return jobOffer.ID, deal.ID
 }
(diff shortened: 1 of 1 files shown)


pkg/solver/controller.go
```diff
@@ -742,6 +754,15 @@ func (controller *SolverController) updateDealState(id string, state uint8) (*da
 		return nil, err
 	}
 
+	// Report validation lost reputation for rejected deals
+	if state == data.GetAgreementStateIndex("MediationRejected") || state == data.GetAgreementStateIndex("ResultsRejected") {
```
narbs91, line 758: [nit] You can omit the validation lost section as we don't currently have validation/mediation implemented yet

narbs91 (approved):

pkg/solver/controller.go
rexdotsh: we can just leave it be? and when we eventually implement it'll work? or should i remove entirely

